PR #1, opened 2021-03-08
Vagrant + bootstrap.sh

danitLevi:
- Adding bootstrap.sh file
- Adding a Vagrantfile for spinning up a development VM
(and enable files sharing , increasing the amount of memory in our
VM,providing HTTP access into our VM , make vagrant run bootstrap.sh for
us )
- Making changes to the `.gitignore` file to ignore so data that Vagrant
creates and ignore nohup.out file

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -127,3 +127,9 @@ dmypy.json
 
 # Pyre type checker
 .pyre/
+
+# Vagrant
+.vagrant/
+
+#nohup.out
+nohup.out
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -1,0 +1,7 @@
+#!/bin/bash -ex
+# The -e option would make our script exit with an error if any command
+# fails while the -x option makes verbosely it output what it does
+
+# Install Pipenv, the -n option makes sudo fail instead of asking for a
+# password if we don’t have sufficient privileges to run it
+sudo -n dnf install -y pipenv
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -1,0 +1,82 @@
+# -*- mode: ruby -*-
+# vi: set ft=ruby :
+
+# All Vagrant configuration is done below. The "2" in Vagrant.configure
+# configures the configuration version (we support older styles for
+# backwards compatibility). Please don't change it unless you know what
+# you're doing.
+Vagrant.configure("2") do |config|
+  # The most common configuration options are documented and commented below.
+  # For a complete reference, please see the online documentation at
+  # https://docs.vagrantup.com.
+
+  # Every Vagrant development environment requires a box. You can search for
+  # boxes at https://vagrantcloud.com/search.
+  config.vm.box = "fedora/32-cloud-base"
+  
+  config.vm.synced_folder ".", "/vagrant", type: "virtualbox"
+  
+  config.vm.provider "virtualbox" do |vb|
+	vb.memory = "1024"
+	end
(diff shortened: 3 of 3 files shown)


.gitignore
```diff
@@ -127,3 +127,9 @@ dmypy.json
 
 # Pyre type checker
 .pyre/
+
+# Vagrant
+.vagrant/
+
+#nohup.out
+nohup.out
```
ifireball, line 135: The last character in a file need to be a newline character - most text editors have a setting that makes the editor take care of that for you,

bootstrap.sh
```diff
@@ -0,0 +1,15 @@
...
+# The -e option would make our script exit with an error if any command
+# fails while the -x option makes verbosely it output what it does
+
+# Install Pipenv, the -n option makes sudo fail instead of asking for a
+# password if we don’t have sufficient privileges to run it
+sudo -n dnf install -y pipenv
+
+cd /vagrant
+# Install dependencies with Pipenv
+pipenv sync --dev
```
ifireball, line 11: I don't think you can do this at this point since you don't have a Pipfile yet.

bootstrap.sh
```diff
@@ -0,0 +1,15 @@
...
+
+# run our app. Nohup and "&" are used to let the setup script finish
+# while our app stays up. The app logs will be collected in nohup.out
+nohup pipenv run python manage.py runserver 0.0.0.0:8000 &
```
ifireball, line 15: missing newline

ifireball (changes requested):
This PR contains parts that will obviously not work if it gets merged. Please make sure you only include working code in your PRs.

bootstrap.sh
```diff
@@ -0,0 +1,11 @@
...
+# The -e option would make our script exit with an error if any command
+# fails while the -x option makes verbosely it output what it does
+
+# Install Pipenv, the -n option makes sudo fail instead of asking for a
+# password if we don’t have sufficient privileges to run it
+sudo -n dnf install -y pipenv
+
+# run our app. Nohup and "&" are used to let the setup script finish
+# while our app stays up. The app logs will be collected in nohup.out
+ nohup pipenv run python manage.py runserver 0.0.0.0:8000 &
```
ifireball, line 11: Given that you don't have a Django application, I don't see how this line would work at this stage.

ifireball:
This PR contains junk commits - please squash your commits together and only leave behind commits that make sense when viewed separately

bootstrap.sh
danitLevi: I tried to run "vagrant up " in my computer and it seem to be working with this line . 
Should I delete it ?

ifireball: Its probably showing you an error message because of this line, or perhaps dumping it into `nohup.out`.

I think you should add this line until you have an actual reason to do it

ifireball (approved):
LGTM, does other team members want to review/comment before we merge this?

liornoy (commented):
In the Vagrantfile maybe delete the comments that are not necessary like lines 1,2 29-81?

rebeccaTubman:
Since the only issue found were the comments that don't impact directly on the performance, may I suggest merging and if needed in the future editing those comments out?

ifireball:
I'm going to merge this for now, because I know everyone else need this to do their work. If you want to make further changes, you can create an issue to track that future work.

PR #3, opened 2021-03-08
Adding CI - Installing Flake8 tool

idokk1:
we will use the “Flake8” tool and set things up so that it scans our code automatically as we send PRs and provides feedback on the PR review screen.
The purpose of CI is to fail when issues exist in the code.

### DIFF
--- a/.github/workflows/flake8.yml
+++ b/.github/workflows/flake8.yml
@@ -1,0 +1,16 @@
+name: Flake8
+on: pull_request
+jobs:
+  flake8:
+    name: Check code with Flake8
+    runs-on: ubuntu-20.04
+    container: fedora:32
+    steps:
+      - name: Install Pipenv and Git
+        run: dnf install -y pipenv git
+      - name: Checkout code
+        uses: actions/checkout@v2
+      - name: Setup environment
+        run: pipenv sync --dev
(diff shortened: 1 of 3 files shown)


.github/workflows/flake8.yml
```diff
@@ -0,0 +1,16 @@
...
+    container: fedora:32
+    steps:
+      - name: Install Pipenv and Git
+        run: dnf install -y pipenv git
+      - name: Checkout code
+        uses: actions/checkout@v2
+      - name: Setup environment
+        run: pipenv sync --dev
+      - name: Run Flake8
+        run: pipenv run flake8 --max-line-length 120
```
liornoy, line 16: Ido I think Barak wants us to add a newline at the end of every file

idokk1: Thanks, I added.

ifireball:
I see this Pr and PR #4 are both adding a `Pipfile` and a `Pipfile.lock`. Once we merge one of these PRs it will cause a severe conflict for the other because you can't edit those files manually to resolve it.

I suggest you choose which one goes in first and rebase the other on top of it in advance while also noting the dependency in the PR description so everyone knows what should be merged first.

PR #6, opened 2021-03-09
Update readme.md

liornoy:
Added the following:
* Badges for: current version, flake8 ci status, maintenance status
* Details about the upcoming features
* Installation, dependencies, and start-up instructions
* Link to the open issues page
* Credit for the Red-Hatters
* License details

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,19 +1,56 @@
 # Shlifim Project - beyond 05 team 3
 
+
 ![shlifim_logo_2](https://user-images.githubusercontent.com/40122521/110316976-2990f880-8014-11eb-8ee8-96c21dd4b29f.png)
 
-Shlifim is the project our group will work on during the course.
+**Description**: Shlifim is the project our group will work on during the course.
 The project will be forum based web application for helping students find answers
 to their homework and assignments and also share summaries.
 
-The develpoment of this project will be in [Vagrant](https://www.vagrantup.com/) environment using [PipEnv](https://github.com/pypa/pipenv) and [Django](https://www.djangoproject.com/)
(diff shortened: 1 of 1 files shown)


danitLevi:
Great readme file .

README.md
```diff
@@ -1,19 +1,56 @@
...
-* Aviv Zafrani
-* Danit Levi
-* Ido Kahlon
-* Rebecca Tubman
-* Lior Noy
+
+## Dependencies
+The dependencies that must be installed for this software to work are:
+* [Vagrant](https://www.vagrantup.com/downloads)
+* [VirtualBox](https://www.virtualbox.org/wiki/Downloads)
```
avivz450, line 33: I think you need to update this part with the dependencies in the `Pipfile`

avivz450 (commented):
Looks good, I think you should also mention the dependencies that are written in the `Pipfile` (flake8, Django, and the version of python)

README.md
liornoy: good point, I debated with myself whether to add those dependencies, but because Vagrant runs a bootstrap script that takes care of the download and installation of Pipenv, flake8, Python, and Django, I think the developer only need to download Vagrant and VirtualBox by himself and the rest of it Vagrant is taking care of.
Do you agree with me or should I still add those to the list in order to inform the developer what tools are being used?

avivz450: Ok, I understand the point. You can keep it like that.

ifireball (approved):
LGTM, do you want to merge this?

danitLevi:
In My opinion you can merge it

PR #20, opened 2021-03-30
Fixed settings.py and urls.py to work for 'home'

rebeccaTubman:
A commit following the last 'Init home app' PR where 'landingpage'
was changed to 'home' but had the files above relate to the previous
name.

modified:   shlifim/settings.py, shlifim/urls.py

### DIFF
--- a/shlifim/settings.py
+++ b/shlifim/settings.py
@@ -37,7 +37,7 @@ INSTALLED_APPS = [
     'django.contrib.sessions',
     'django.contrib.messages',
     'django.contrib.staticfiles',
-    'landingpage.apps.LandingpageConfig',
+    'home.apps.HomeConfig',
 ]
 
 MIDDLEWARE = [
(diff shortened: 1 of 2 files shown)


liornoy:
Great fix

ifireball:
I'm seeing the URL from the landing page being removed, but I'm not seeing a URL for the new home app beong added. Where is it?

rebeccaTubman:
I didn't added one because we haven't created a view for home app yet.

liornoy:
> I'm seeing the URL from the landing page being removed, but I'm not seeing a URL for the new home app being added. Where is it?

We will make different PR for it with the needed changes to views and add templates and urls

danitLevi:
> > I'm seeing the URL from the landing page being removed, but I'm not seeing a URL for the new home app being added. Where is it?
> 
> We will make different PR for it with the needed changes to views and add templates and urls

I Agree . I think it's okay if at this point when entering our web the default Django success message will be displayed and we will add our new home page in a separate PR in the future and update the url file .

ifireball (approved):

PR #21, opened 2021-03-30
Created User Model and Migration files

### LINKED ISSUE #13: Create User Model and Migrations
Create a DB for Users following the table in the ERD which includes: 
- user ID (PK)
- user name
- password
- registration date
- email address
- gender
- is blocked
- is admin

Assigned to: @rebeccaTubman

rebeccaTubman:
Following the ERD for Shlifim and using imported 'User' from
'django.contrib.auth.models', the Profile model will be used as the
base user in the Shlifim website.

Where the Fields:
user name = username (Imported from User)
password = password (Imported from User)
registration date = date_joined (Imported from User)
email address = email (Imported from User)
gender = gender
is blocked = is blocked
is admin = is_superuser & is_stuff (Imported from User)

- Modified admin.py to view the database in the admin page.
- Created 0002_test_data.py to text initial data for the Profile model.

modified:   home/admin.py, home/models.py
new file:   home/migrations/0001_initial.py,
            home/migrations/0002_test_data.py

fix #13


danitLevi:
LGTM

liornoy:
image

Looking at the _str_ function of the Profile model, at first, I thought it's great and informative, but after seeing how it is displayed by Django's admin page maybe it will be better to short it to display the profile's username only

rebeccaTubman:
Good idea. I made the str function to see all the information in the shell for testing.
Since it'll be used for the Django's admin page I'll change it to username only.

home/migrations/0002_test_data.py
```diff
@@ -0,0 +1,42 @@
...
+            ('Danit', 'F'),
+            ('Aviv', 'M'),
+        ]
+
+        with transaction.atomic():
+            # Test: Create a User and Profile from a User via username.
+            for username, password, email in user_test_data:
+                User(username=username, password=password, email=email).save()
+            for username, gender in profile_test_data:
+                Profile(user=User.objects.get(username=username), gender=gender).save()
```
simzacks, line 29: While this will work, what I would suggest is to add the profile data into the user_test_data list, and process them together:
For example:
```
user_test_data = [
            ('Rebecca', 'RebeccaRebecca', 'Rebecca@gmail.com', 'F'),
            ('Danit', 'DanitDanit', 'Danit@gmail.com', 'F'),
            ('Aviv', 'AvivAviv', 'Aviv@gmail.com', 'M'),
            ('Ido', 'IdoIdo', 'Ido@gmail.com', None),
        ]
        with transaction.atomic():
            # Test: Create a User and Profile from a User via username.
            for username, password, email, gender in user_test_data:
                user = User(username=username, password=password, email=email).save()
                if gender:
                    Profile(user=user, gender=gender).save()

```

home/models.py
```diff
@@ -1,3 +1,32 @@
...
+               Raw passwords can be arbitrarily long and can contain any character.
+    date_joined - A datetime designating when the account was created. Is set to the current date/time by default
+        when the account is created.
+    email - Email address.
+    is_superuser - Boolean. Designates that this user has all permissions without explicitly assigning them.
+    Added Fields:
+    gender - Default gender is set to 'Unspecified', A user can choose to specifie its gender to Male/Female.
+    is blocked - Boolean. Designates that this user won't be able to login until an admin unblockes him.
+    '''
+    user = models.OneToOneField(User, on_delete=models.CASCADE, primary_key=True)
```
simzacks, line 27: While this will work, the best practice is to refer to the AUTH_USER_MODEL.
See: https://learndjango.com/tutorials/django-best-practices-referencing-user-model

home/migrations/0002_test_data.py
rebeccaTubman: Thank you for reviewing my PR. I'll change both files as you suggested.

home/migrations/0002_test_data.py
```diff
@@ -0,0 +1,36 @@
...
+            ('Rebecca', 'RebeccaRebecca', 'Rebecca@gmail.com', 'F'),
+            ('Danit', 'DanitDanit', 'Danit@gmail.com', 'F'),
+            ('Aviv', 'AvivAviv', 'Aviv@gmail.com', 'M'),
+            ('Ido', 'IdoIdo', 'Ido@gmail.com', None),
+        ]
+
+        with transaction.atomic():
+            # Test: Create a User and Profile from a User via username.
+            for username, password, email, gender in user_test_data:
+                user = User(username=username, password=password, email=email).save()
```
simzacks, line 21: https://docs.djangoproject.com/en/3.1/topics/auth/default/
```
>>> from django.contrib.auth.models import User
>>> user = User.objects.create_user('john', 'lennon@thebeatles.com', 'johnpassword')
```
I believe that using the create_user function enccrypts the passord properly, while just using save() does not.

home/migrations/0002_test_data.py
```diff
@@ -0,0 +1,36 @@
...
+                if gender:
+                    Profile(user=user, gender=gender).save()
+                else:
+                    Profile(user=user).save()
```
simzacks, line 25: If you want to create the profile whether there is gender or not, you can remove the if statement.
`Profile(user=user, gender=gender).save()`
when gender is None, it the same as 
`Profile(user=user).save()`

simzacks (changes requested):

home/migrations/0002_test_data.py
rebeccaTubman: Thank you. 
I've made all the changes you suggested.

simzacks (approved):

home/models.py
```diff
@@ -1,3 +1,32 @@
...
+    date_joined - A datetime designating when the account was created. Is set to the current date/time by default
+        when the account is created.
+    email - Email address.
+    is_superuser - Boolean. Designates that this user has all permissions without explicitly assigning them.
+    Added Fields:
+    gender - Default gender is set to 'Unspecified', A user can choose to specifie its gender to Male/Female.
+    is blocked - Boolean. Designates that this user won't be able to login until an admin unblockes him.
+    '''
+    user = models.OneToOneField(settings.AUTH_USER_MODEL, on_delete=models.CASCADE, primary_key=True)
+    gender = models.CharField(max_length=1, choices=Gender.choices, default=Gender.UNSPECIFIED, blank=True, null=True)
```
ifireball, line 28: If you have a default value, why do you need to allow it to be blank or null?

ifireball (changes requested):

home/models.py
rebeccaTubman: I used the exact same definition as the one in the presentation we were shown on Sunday.
Is it wrong?

ifireball: IMO yes. Blank and Null almost never make sense being both true, and blank is probably not even relevant for CharField (AFAIK a single char cannot be a blank string).

Blank means the field can contain a blank string - it only makes sens for strings and in many cases you don't want to allow it, unless the field is for something like an optional description text.

NULL means the field does not contain a value at all (Like a NULL pointer). Sometimes NULLs in the database are used to indicate values that were not specified by the used (As opposed to blank which can mean that the value was specified but was an empty string), but in this case you already designated an UNSPECIFIED value, so also allowing NULL as well does not man much sense.

If you do want to allow for NULLs, you will have to make sure you check for it everywhere in order to avoid null pointer de-referencing, so in general its better not to allow it...

rebeccaTubman: I'll change it now.
I think this change will collide with how the Profile is created in data_test_0002.
 `Profile(user=user, gender=gender).save()`
so I'll change it to `if gender ... else`

ifireball: See comment by @simzacks , the "if" statement is probably not needed there.

rebeccaTubman: Without the if it made an error, this is why I added the if else.

home/migrations/0002_test_data.py
ifireball: As discussed elsewhere, it seems that having `gender=None`is not quite the same as not setting it (The first tries to explicitly set a NULL value on the field while the 2nd uses the defined default).

ifireball (approved):

PR #24, opened 2021-03-31
Add subject, subsubjcet, book models

### LINKED ISSUE #16: Create Subject, Sub Subject and Book Model and Test Migration
Subject, Sub subject, and book models will be added to models.py following this ERD design: 
https://lucid.app/publicSegments/view/f1d0a9f5-f835-4cb8-9080-1adadcacc922/image.png

Also, create a Migration file for creating test-data

Assigned to: @liornoy

liornoy:
* Add Subject, Subsubject, Books - models to model.py
* Register the models in admin.py
* Update migration to include the new models. (init + test data)

fix #16 , related to #27


danitLevi:
LGTM

home/models.py
```diff
@@ -1,3 +1,24 @@
...
+    def __str__(self):
+        return self.sub_subject_name
+
+
+class Book(models.Model):
+    book_name = models.CharField(max_length=100)
+    related_subject = models.ForeignKey(Subject, on_delete=models.CASCADE)
+
+    def __str__(self):
+        return self.book_name
```
simzacks, line 24: A couple of comments. 
If this is what you actually want, then it is ok as is.
* This structure only allows for one level of sub_subject
* Book can only be tied to a subject and not a sub_subject
* If you don't want to attach a sub subject to a book, what do you plan on doing with the table?
* Can a book have multiple subjects? (according to this design, it can't)

liornoy: Yes, we planned on having sub-subjects for each subject. (just one level)
Books will be tied to subject, and only one.
And the user's questions in the app will be categorized in certain subject, and sub-subject, with the option to add a book reference with the same subject.

simzacks (approved):

ifireball (changes requested):
If we merge this we cannot merge MR #21, and vice versa, which one do you want to go in first?

Also, it does not make much sense to introduce model classes without introducing the necessary migration files. This is essentially introducing broken code, because the classes cannot be used without running the equivalent migrations first.

danitLevi:
I think you should merge Pr #21 first

ifireball (changes requested):
Please do not modify merged migration files, add new ones!

liornoy:
> Please do not modify merged migration files, add new ones!

Fixed it

home/migrations/0004_test_data_book_subsubject_subject.py
```diff
@@ -0,0 +1,84 @@
...
+            ('All My Sons', 'Literature'),
+            ('1948', 'Literature'),
+            ('antigona', 'Literature'),
+            ('Sapiens: A Brief History of Humankind', 'History'),
+            ('Modern Hebrew for Beginners', 'Hebrew'),
+            ('Benny-Goren', 'Math'),
+            ('Fundamentals of Physics', 'Physics')
+        ]
+
+        with transaction.atomic():
```
ifireball, line 69: This is quite an inefficient way to load this data, since you are running a database query fro every single Sub_Subject and Book record you're adding.

This could potentially be optimized by nesting that test data structures, so that you can have an outer loop that added the Subject records and an inner loop that added the Sub_Subject and Book records.

In any case will merge for now in the interest of time.

home/models.py
```diff
@@ -30,3 +30,31 @@ class Profile(models.Model):
...
+
+
+class Subject(models.Model):
+    subject_name = models.CharField(max_length=100, unique=True)
+
+    def __str__(self):
+        return self.subject_name
+
+
+class Sub_Subject(models.Model):
```
ifireball, line 42: The standard for Python class names is `CamelCase`. `snake_case` is used for function, method and property names. I don't think I've seen such a mixed name form used anywhere...

ifireball (approved):

Issue #27, opened 2021-04-02
Issues regarding the presentation of models on the admin page


liornoy:
Looking at the models through the admin page, I noticed that:

#### 1. Uniting the User and the Profile models to one page on the admin page
There is a different page for viewing users and viewing profiles which in my opinion should be on the same page because the profile model is an extension for the user model, and has one on one relation.

#### 2. How models are being displayed on the admin page: 
In models such 'Book' 'Question', 'Sub-subject', 'Answer' the admin page is showing by default only one column of information, for example for Books it shows only the book's name, but it would be better if it showed the related subject in the same row.

### The solutions:
1. Found [this tutorial](https://simpleisbetterthancomplex.com/tutorial/2016/11/23/how-to-add-user-profile-to-django-admin.html) that shows how to display the Profile information inside the built-in user model page in admin.
2.  Found [this StackOverflow thread](https://stackoverflow.com/questions/163823/can-list-display-in-a-django-modeladmin-display-attributes-of-foreignkey-field) discussing ModelAdmin classes that enable you to add 'list_display' field which customize the way the admin page displays the data of our models. This way the admin page can be more informative. 

Example of the Book model page on the admin page before and after the change 

Before:
![image](https://user-images.githubusercontent.com/40122521/113416668-7b862d80-93ca-11eb-870d-c4da0dd08236.png)

After:
![image](https://user-images.githubusercontent.com/40122521/113416639-6c9f7b00-93ca-11eb-9ba8-ba0e0ec4d3c3.png)

idokk1:
the StackOverflow solution seems pretty simple, I think we should apply it.

PR #34, opened 2021-04-06
Adding Question model

### LINKED ISSUE #19: Create Question Model and Migrations
Create a DB table for questions following the table in the ERD which includes:
* question id(PK)
* user id(FK)
* title
* content
* publish date
* subject id(FK)
* grade
* sub subject id (FK)
 * book id (FK)
* book page 
* is edited

danitLevi:
- Updating modles.py file:
	* add Grade class which contain grade choices (Seventh Grade until Twelfth grade) 
	* add Question class based on the ERD 
- Updating admin.py file in order to view questions data in admin page
- update migrations files

fix #19 
related to #27

*  I also included the code from the unmerged PR Lior opened (#24 ) because I needed  subject, sub-subject and book models


liornoy:
It looks great.
Maybe add some more questions to the test_data so the initial database will be fuller,
and so we will have content to play with when designing the views later.

Preferably the questions will include different authors and different publish dates,
with/without books etc.
Also in home/migrations/0002_test_data.py. Looking at: 
```
for username, title, content, publish_date, subject_id, sub_subject_id, \
grade, book_id, book_page, is_edited in question_test_data: 
```

I think the question_test_data list items will be more readable if you used the object's name instead of id. this way instead of: 
```
('Danit', 'question from math course', 'Hey , Can someome help me solve the equation 1+1',
             timezone.now(), 9, 1, '7', 1, 10, True),
```

You'll have something like:
 ```

('Danit', 'question from math course', 'Hey , Can someome help me solve the equation 1+1',
             timezone.now(), 'Math', 'Algebra', '7', 'Yoel Geva', 10, True),
 ```

idokk1:
image
Hi , 
I think maybe in you __str__ function you should return on the content and ID. The date is looks bad when I show it in the admin page.

idokk1:
maybe change the str function to : 
```
 def __str__(self):
```
and then it will look like this : 
image

danitLevi:
@idokk1 
I updated admin.py file and added "QuestionAdmin" class , so in the admin page the question Shown in a table with the following columns: id, title , publish date , subject, and not the as presented In the str function. 
(I followed Lior's suggestion in issue #27 ) .
It worked fine for me. Maybe you forgot to update admin.py file?
In addition , Let me know if you want me to change something in QuestionAdmin class .

idokk1:
I use the str function when I want to show the question as a column in the Answer admin page: 
image

liornoy:
> Good idea @liornoy .
> I will add some more questions to the test_data.
> 
> In addition , I thought we wanted to map objects by id.
> I can change it, but if we will map objects by their names we should also change the Subject, Sub-subject, and Book models and convert their name fields into unique fields ( using unique=True )

I changed my PR as you suggested. thank you

danitLevi:
I will change it as you suggested thank you .

danitLevi:
> maybe change the str function to :
> 
> ```
>  def __str__(self):
>         return f"Question #{self.id} : {self.title} ( {self.publish_date.strftime('%d/%m/%Y %H:%M')} )"
> ```
> 
> and then it will look like this :
> image

I will change it as you suggested thank you @idokk1  .

danitLevi:
@liornoy @idokk1 
I did the required changes. 
Let me know if you want me to do more changes.

idokk1:
LGTM

ifireball (changes requested):
Merged migrations files should not be modifoed, instead, new files must be atted.
Since all of you are introducing models into the same Django app, all the model modification MRs conflict with one another and more be based on each other.

ifireball:
Please resolve the merge conflict - one possible reason for it is that this MR is attempting to change the 002 migration file (as always, migration files shouldn't be changed once they are merged)

danitLevi:
> Please resolve the merge conflict - one possible reason for it is that this MR is attempting to change the 002 migration file (as always, migration files shouldn't be changed once they are merged)

@ifireball Done

ifireball (approved):

PR #37, opened 2021-04-08
Change home's admin.py file to unite profile and user

liornoy:
This change will move the presentation of the profile into
the same page as the User objects in the admin page

Related to #27


ifireball:
Since the main use of the admin UI or now is to present your model structure in the coming demo, Id recommend against making this change.

PR #43, opened 2021-04-11
Change fields in Question models

### LINKED ISSUE #42: Fix NOT NULL Constraint in Question model
In the home app, models.py file. the Question model, the 'book' field is set to:
```
book = models.ForeignKey(Book, on_delete=models.CASCADE, blank=True)
```
When creating a question without a book Django throws: "NOT NULL constraint failed"
should add null=true to solve this

The same goes for sub_subject field

liornoy:
Added null=True to sub_subject, and book fields
to solve the "NOT NULL constraint" error when creating
 a new question without sub-subject or book

fix #42


avivz450:
Good update.
After going over the model.py code again, it seems that there are no more such cases in the other models

ifireball (approved):
The PR title could be improved, changing something that couldn't be null before into supporting being null, is essentially making it optional. The title could have indicated that.

PR #45, opened 2021-04-13
Display question feature

### LINKED ISSUE #44: Displaying certain question page feature
This feature is for the page that opens after you click certain questions on the "Explore" page.
The page will display the question, and underneath all the existing answers.
Also, if this page is opened by the user who wrote the question he will have the options to delete and edit.

### Tasks:
1. Add the needed function/class to views.py
2. Add the route to the urls,py
3. Create the needed HTML file
4. Add the needed tests for the feature

### The "prototype" design: 
image

### LINKED ISSUE #69: Write tests for display-certain-question feature

liornoy:
This PR is for the display question feature.
it includes changes in the URLs, views, models, and a new HTML file for this feature.
also added CSS file for all the needed classes.

Also added tests to tests.py using Pytest

fix #44 
fix #69


home/templates/home/question_detail.html
```diff
@@ -0,0 +1,64 @@
...
+	</div>
+</div>
+<div class="answers-container">
+	{% for answer in answers %}
+			<div class="answer-display ">
+				<p class = "left">{{ answer.content }}</p>
+				<div class="answer-meta-data right pl-4">
+					<p>by <a class="mr-2" style="color:#6FB5EE" href="#"> <b>{{ answer.profile }}</b></a></p>
+					<small class="text-muted date">{{ object.publish_date|date:"F d, Y" }}</small>
+					<p style="text-align:center"><br><br>{{ answer.getSumOfLikes }} </p>
```
danitLevi, line 54: Good job Lior !
I think it would look better to show the number of likes and  the number of dislikes (next to the thumbs up and thumbs down icons) ,instead of showing the subtraction between them

home/templates/home/question_detail.html
```diff
@@ -0,0 +1,64 @@
...
+					<button style="font-size:24px"><i class="fa fa-thumbs-o-down"></i></button>
+					<button style="font-size:24px"><i class="fa fa-thumbs-o-up"></i></button>
```
danitLevi, line 56: Good job Lior !
I think it would look better to show the number of likes and the number of dislikes (next to the thumbs up and thumbs down icons) ,instead of showing the subtraction between them

liornoy: thanks, at first I thought to show the subtraction because this number dictates the display order of the answers when clicking the "votes" button. Nevertheless, I think showing both likes and dislikes can work great too.
I'll create the template for it and we'll compare and decide what to go with.

danitLevi:
I think it's a better practice to add the "status/in progress" label instead of writing "wip" in the PR headline.

idokk1:
LGTM

home/models.py
```diff
@@ -83,11 +83,16 @@ class Question(models.Model):
     is_edited = models.BooleanField(default=False)
 
     class Meta:
-        ordering = ['-publish_date']  # ordered by publish_date
+        ordering = ['-publish_date']  # default ordered by publish_date
 
     def __str__(self):
         return f"Question #{self.id} : {self.title} ( {self.publish_date.strftime('%d/%m/%Y %H:%M')} )"
 
+    @classmethod
```
ifireball, line 91: If you need a method that returns an information that i relevant for a particular questions oyu can adde a regular method instead of a class method, that way you can avoid the extra DB query. The same comment seem to be relevant to all the methods you've added here.

home/views.py
```diff
@@ -7,3 +9,37 @@ def about(request):
 
 def landingpage(request):
     return render(request, 'home/landingpage.html')
+
+
+class QuestionDetailView(DetailView):
+    model = Question
+
+    def get_context_data(self, **kwargs):
```
ifireball, line 17: I think this is not a good function design - it has many very different things included in it, think about how difficult it would be to write tests for it.

I think it should be broken apart and most stuff in it moved out to model functions.

home/views.py
```diff
@@ -7,3 +9,37 @@ def about(request):
 
 def landingpage(request):
     return render(request, 'home/landingpage.html')
+
+
+class QuestionDetailView(DetailView):
```
ifireball, line 14: Not sure what is the usefulnes of having class-based views here, IMO those seem to introduce extra code builerplate without visible benefits.

home/templates/home/question_detail.html
```diff
@@ -0,0 +1,242 @@
+{% extends "home/base.html" %}
+{% load static %}
+{% block content %}
+
+
+<style>
```
ifireball, line 6: I'd recommend having a shared, static CSS file - its better for browser-side caching.

home/tests.py
```diff
@@ -1,3 +1,51 @@
-# from django.test import TestCase
+from django.test import TestCase
+from .models import Question, Answer
+from django.test import Client
+from django.urls import reverse
 
-# Create your tests here.
+
+class QuestionModelTests(TestCase):
```
ifireball, line 7: You don't need to define a test class and certainly nit inherit from TestCase. This is how you use [unittest][1] while we instructed you to use [PyTest][2]

[1]: https://docs.python.org/3/library/unittest.html
[2]: https://docs.pytest.org/en/6.2.x/

ifireball (changes requested):

home/templates/home/question_detail.html
liornoy: We planned to move all styles to shard CSS file after everyone will merge their templates, so won't be conflicts in the shared file

ifireball: You'll get code duplication that way... If some parts of your UI have separate, distinct styles, you can put them in separate CSS files.

liornoy: I moved all the CSS to different file

liornoy:
@ifireball , @AviadP the PR is ready for re-review, please

ifireball:
You need to rebase this PR - it has conflicts - please don't leave it up to me to tell you this, you can look at the PR screen and scroll down just like I can.

home/models.py
liornoy: > You need to rebase this PR - it has conflicts - please don't leave it up to me to tell you this, you can look at the PR screen and scroll down just like I can.

Sorry I didn't notice the new PR install Pytest was merged.
I fixed my PR and is ready to review

home/views.py
```diff
@@ -7,3 +8,30 @@ def about(request):
 
 def landingpage(request):
     return render(request, 'home/landingpage.html')
+
+
```
ifireball, line 12: Not seeing any tests for this view

home/views.py
```diff
@@ -7,3 +8,30 @@ def about(request):
...
+def create_display_question_context(question, sortAnsBy):
+    context = {}
+    context['question'] = question
+
+    answers = question.get_answers_feed(sortAnsBy)
+    context['answers'] = answers
+
+    # Get count of answers string
+    answersCount = answers.count()
+    context['answersCountStr'] = str(answersCount) + ' Answer' + ('s' if answersCount > 1 else '')
```
ifireball, line 29: This can be a custom template function, though IIRC the templating engine already has nice functions for dealing with plural v.s. sigular value titles.

home/tests.py
```diff
@@ -1,3 +1,83 @@
+from .models import Question, Answer, Profile
+from django.test import Client
+from django.urls import reverse
+from datetime import datetime
+from django.db.models.query import QuerySet
+import pytest
 
-def test_first():
-    assert True
+
```
ifireball, line 8: Since you're going to end up with a very big tests file, perhaps you want to split it up ahead of time, or at least use classes to group related tests together.

ifireball (changes requested):
see comments on the code - generally I think this PR has too much stuff in it, so it takes a lot of time to review

home/views.py
liornoy: I don't understand what you mean you don't see tests for this view.
Every test exists because I need it for the view:
* The answer feed function which queries the answers to display,
* The thumb up / down functions
* The question title I use in the view
* The HTTP status response
All of them don't have a purpose without the view display question I made.
What am I missing?

home/views.py
liornoy: Ok, fixed

home/tests.py
liornoy: Classes were added to group-related tests.

home/tests.py
```diff
@@ -1,3 +1,81 @@
...
+            assert(question.get_question_title() == "Math-question from math course")
+
+    class TestHTMLRelated:
+        @pytest.mark.django_db
+        def test_display_question_page_url(self):
+            """
+            This test checks the returned status for routing to display-question feature
+            """
+            url = reverse('question-detail', args=[1])
+            c = Client()
```
ifireball, line 79: Use the `client` fixture that the `pytest-django` plugin provides for you. We've shown at least two examples of using it in class.

Also you should probably test more then just the HTTP status code, you should at least verify that the right template had bee used and the context passed to it contains the right contents

home/views.py
ifireball: I think the method you have above that checks the view through the client wasn't there the last time I looked

home/views.py
```diff
@@ -7,3 +8,29 @@ def about(request):
...
+
+
+def displayQuestion(request, **kwargs):
+    question = Question.objects.all().get(id=kwargs['pk'])
+    sortAnsBy = request.GET["sortanswersby"] if 'sortanswersby' in request.GET else ''
+    context = create_display_question_context(question, sortAnsBy)
+    return render(request, 'home/question_detail.html', context)
+
+
+def create_display_question_context(question, sortAnsBy):
```
ifireball, line 20: why is this function still here? ignoring my comments about it will not make me forget them.

ifireball (changes requested):
see comments in the code

home/views.py
liornoy: I didn't ignore it, I did change and moved some functions to the models.py. Seeing that it's still not good enough, I moved all the context-making login to models.

home/tests.py
liornoy: OK fixed

home/views.py
liornoy: fixed

home/tests.py
liornoy: fixed

liornoy:
> see comments in the code
@ifireball @AviadP
Fixed all the comments and ready for review

home/views.py
liornoy: changed to FBV

home/tests.py
```diff
@@ -96,3 +95,97 @@ def question_tag_test_data(question_test_data):
...
+        @pytest.mark.django_db
+        def test_response_context(self, response):
+            '''
+            Testing if the context passed to the view contains the right contents
+            '''
+            question = Question.objects.get(id=2)
+            context = question.make_context('')
+            checkList = ['question', 'answers', 'answersCount', 'tags', 'title']
+            for check in checkList:
+                assert str(response.context[check]) == str(context[check])
```
ifireball, line 191: this is checking the code with itself, its a useless test

if you want to ensure that function A calls function B and returns a value that is based on the value returned from function B, you need to use a mockup of function B and a sentinel value.

Mockups and sentinel values an be created with the unitest.mock library, and be injected into the application via the `monkeypatch` fixture.

But since you have very tight cntrol over the test data in the system, you can avoid doing any of that and just compare the data in the context you got to a hard-wired set of expected values.

ifireball (changes requested):

home/models.py
```diff
@@ -104,6 +104,39 @@ def add_tags_to_question(self, tags):
...
+        if filterType == 'date' or filterType == '':
+            answers = answers.order_by('-publish_date')
+        elif filterType == 'votes':
+            answers = answers.order_by('-likes_count')
+        return answers
+
+    def get_question_title(self):
+        return str(self.subject) + '-' + self.title
+
+    def make_context(self, sortAnsBy):
```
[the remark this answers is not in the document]
liornoy: Issue fixed, made it dictionary creation expression

home/tests.py
liornoy: Fixed, I compare it now with a hard-wired set of expected values.
Please note that I use string.startswith function because the question date is created to Now() time.
I know it's problematic, but I will change it after we will have more test data.
I opened an issue: #84 for it.

ifireball (approved):

PR #64, opened 2021-04-15
Adding DB test for adding new answer and new profile

### LINKED ISSUE #73: Create Answer tests
Creating Answer model tests (before Edward class).
After all the features will merge and we'll have a view.py , I will add more tests.

idokk1:
adding test for creating new user, profile and answer to the DB.
create PyTest YML file to run  the application's tests with PyTest on PR. (like what we are familiar with flake8)
Pipfile, Pipfile.lock modified by Django while installing PyTest .

made this PR due to Edward's request on Slack. 
fix #73

### DIFF
--- a/home/models.py
+++ b/home/models.py
@@ -150,12 +150,10 @@ class Answer(models.Model):
     def set_is_edited(self, newVal):
         self.is_edited = newVal
 
-# =======
-#     @classmethod
-#     def get_answers_by_date(cls):
-#         return cls.objects.order_by('-publish_date')
-# >>>>>>> e490b6c (Add create method to profile model)
-#
+    @classmethod
+    def get_answers_by_date(cls):
+        return cls.objects.order_by('-publish_date')
+
 
 class Tag(models.Model):
     tag_name = models.CharField(max_length=20)
--- a/home/models.py
+++ b/home/models.py
@@ -1,6 +1,7 @@
 from django.conf import settings
 from django.db import models
 from django.utils import timezone
+from django.contrib.auth.models import User
 
 
 class Gender(models.TextChoices):
@@ -31,6 +32,13 @@ class Profile(models.Model):
 
     def __str__(self):
         return '{self.user.username}'.format(self=self)
+
+    @classmethod
+    def create(cls, username, password, email, gender='U', is_blocked=False):
+        user = User.objects.create_user(username=username, password=password, email=email)
+        profile = cls(user=user, gender=gender, is_blocked=is_blocked)
+        profile.save()
+        return profile
 
 
 class Subject(models.Model):
@@ -142,6 +150,12 @@ class Answer(models.Model):
     def set_is_edited(self, newVal):
         self.is_edited = newVal
 
+# =======
+#     @classmethod
+#     def get_answers_by_date(cls):
+#         return cls.objects.order_by('-publish_date')
+# >>>>>>> e490b6c (Add create method to profile model)
+#
 
 class Tag(models.Model):
     tag_name = models.CharField(max_length=20)
@@ -150,7 +164,7 @@ class Tag(models.Model):
     def __str__(self):
         return self.tag_name
 
-    # The function returnss all tags ordered by name tag.
+    # The function returns all tags ordered by name tag.
     @classmethod
     def tags_feed(cls, search=''):
         return cls.objects.filter(tag_name__contains=search).order_by('tag_name')
@@ -164,7 +178,7 @@ class Question_Tag(models.Model):
         constraints = [
             models.UniqueConstraint(
                 fields=['question', 'tag'], name='question_tag')
-        ]
+                      ]
 
     def __str__(self):
         return f"{self.question}, Tag : {self.tag}"
--- a/home/tests.py
+++ b/home/tests.py
@@ -1,4 +1,4 @@
-from home.models import Profile, Subject, Question, Tag, Answer, Question_Tag
+from home.models import Profile, Subject, Question, Tag, Question_Tag, Answer
 from django.contrib.auth.models import User
 from django.utils import timezone
 from datetime import datetime
@@ -205,3 +205,35 @@ class TestTagsPage:
         url = reverse('tags')
         response = client.get(url)
         return response
+
+
+@pytest.fixture
+def profile():
+    profile = Profile.create(username="test_user", password="testtest", email="test@test.com")
+    return profile
+
+
+@pytest.mark.django_db
+def test_default_params_profile(profile):
+    assert profile.user.id >= 0  # check if user saved successfully and get an id
+    assert profile.gender == 'U'  # default gender is U
+    assert profile.is_blocked is False
+
+
+@pytest.mark.django_db
+def test_default_params_answer(profile):
+    question = Question.objects.first()
+    answer = Answer(profile=profile, question=question, content='Test')
+    assert answer.likes_count == 0
+    assert answer.dislikes_count == 0
+    assert answer.is_edited is False
+
+
+@pytest.mark.django_db
+def test_get_answers():
+    out = Answer.get_answers_by_date()
+    assert all(isinstance(a, Answer) for a in out)
+    assert set([
+        (2, 2, 'IDK'),
+        (1, 1, 'pretty sure its 2 but I suggesting you to check with another resources '),
+        ]).issubset(list(out.values_list('profile', 'question', 'content')))


avivz450:
I think you should set Edward as a reviewer because of what he wrote in Slack

home/tests/test_add_answer_profile.py
```diff
@@ -0,0 +1,28 @@
...
+
+
+@pytest.mark.django_db
+class TestModels:
+
+    def test_add_profile(self):
+        profile = self.get_profile()
+        assert profile.user.id >= 0  # check if user saved successfully and get an id
+        assert profile.gender == 'U'  # default gender is U
+        assert isinstance(profile, Profile)  # check if the profile is a instance of Profile
```
EdDev, line 13: Will these pass even if you remove the `profile.save()`?
It is unclear if the objects here are in sync with the db or not (I would expect a "fetch" operation if the intention is to cover the db as well).

In addition, I do not see a point in checking the type of profile, it is created as an instance of it explicitly on line 26.
Same for the other cases.

EdDev (commented):
Very nice to see this land before the session we had today.

home/tests/test_add_answer_profile.py
idokk1: > Will these pass even if you remove the `profile.save()?`
No, I check that the command success.

I added a tests on DB , I hope that's what you meant.

home/tests.py
```diff
@@ -1,3 +1,34 @@
...
+from home.models import Profile, Question, Answer
+import pytest
 
-# Create your tests here.
+
+@pytest.mark.django_db
+class TestModels:
+
+    def test_add_profile(self):
+        profile = self.get_profile()
```
ifireball, line 10: please turn `get_profile` into a fixture (it could just be called `profile` then)

home/tests.py
```diff
@@ -1,3 +1,34 @@
...
+
+    def test_add_answer(self):
+        profile = self.get_profile()
+        question = Question.objects.first()
+        if profile and question:
+            answer = Answer(profile=profile, question=question, content='Test')
+            answer.save()
+            assert answer.id >= 0  # check if answer saved successfully and get an id
+
+    def get_profile(self):
```
ifireball, line 22: it would be much more useful to move the lines below into a classmethod on one of the models, it would also make more sense to test it then!

home/tests.py
```diff
@@ -1,3 +1,34 @@
-# from django.test import TestCase
+from django.contrib.auth.models import User
+from home.models import Profile, Question, Answer
+import pytest
 
-# Create your tests here.
+
+@pytest.mark.django_db
+class TestModels:
```
ifireball, line 7: You don;t need to put tests in classes when using pytest - if you do, its better to give the classes some speficig names that indicate which tests do they group together - and it doesn;t ususally make sense t group all the model tests together in the same class...

ifireball (changes requested):

idokk1:
@ifireball I fixed what you recommended

home/models.py
```diff
@@ -102,6 +102,15 @@ class Answer(models.Model):
     def __str__(self):
         return self.content
 
+    @classmethod
+    def get_Answers_by_date(cls):
+        return cls.objects.order_by('-publish_date')
+
+    @classmethod
+    def test_Save_Answer(cls, answer):
```
ifireball, line 110: Is this supposed to be a test? Then it does not belong here..

I really don;t understand what kind of logic are you trying to encapsulate here

home/tests.py
```diff
@@ -1,3 +1,35 @@
-# from django.test import TestCase
+from django.contrib.auth.models import User
+from home.models import Profile, Question, Answer
+import pytest
 
-# Create your tests here.
+
+@pytest.fixture
+def profile():
```
ifireball, line 7: Considering that adding the user and profile records is a complex operation, are you sure you don't need a model method for it?

Also consider that you probably want to be adding the two DB object in a single transaction

ifireball (changes requested):

home/models.py
```diff
@@ -32,6 +33,12 @@ class Profile(models.Model):
     def __str__(self):
         return '{self.user.username}'.format(self=self)
 
+    @classmethod
+    def create(cls, username, password, email, gender='U', is_blocked=False):
```
ifireball, line 37: The DB operations below probably aught to be done in a transaction together

Why arn't you saving the profile object?

home/tests.py
```diff
@@ -1,3 +1,40 @@
-# from django.test import TestCase
+from django.contrib.auth.models import User
+from home.models import Profile, Question, Answer
+import pytest
 
-# Create your tests here.
+
+@pytest.mark.django_db
+def test_Add_Profile():
```
ifireball, line 7: If you have a profile creation class method, what is the poing of rewriting profile creation code is a part of a test? What is this testing exactly?

home/tests.py
```diff
@@ -1,3 +1,40 @@
...
+def profile():
+    profile = Profile.create(username="test_user", password="testtest", email="test@test.com")
+    profile.save()
+    return profile
+
+
+@pytest.mark.django_db
+def test_Add_Answer(profile):
+    question = Question.objects.first()
+    if profile and question:
```
ifireball, line 26: does this "if" make sense here? if the test DB does not contain the expected objects, should we just blindly skip testing everything?

home/tests.py
```diff
@@ -1,3 +1,40 @@
...
+        answer = Answer(profile=profile, question=question, content='Test')
```
ifireball, line 27: What is the logic theses asserts are testing? Are they desting the default values for the attributes blow? The test name does not help in understanding that....

home/models.py
```diff
@@ -102,6 +109,10 @@ class Answer(models.Model):
     def __str__(self):
         return self.content
 
+    @classmethod
+    def get_Answers_by_date(cls):
```
ifireball, line 113: please stick to the usual Python coding conventions of having method names be fully lower case

ifireball (changes requested):

home/tests.py
```diff
@@ -1,3 +1,35 @@
-# from django.test import TestCase
+from home.models import Profile, Question, Answer
+import pytest
 
-# Create your tests here.
+
+@pytest.fixture
+def profile():
+    profile = Profile.create(username="test_user", password="testtest", email="test@test.com")
+    profile.save()
```
liornoy, line 8: You did profile.save() inside the Profile.create already

home/models.py
```diff
@@ -102,6 +110,10 @@ class Answer(models.Model):
     def __str__(self):
         return self.content
 
+    @classmethod
+    def get_answers_by_date(cls):
+        return cls.objects.order_by('-publish_date')
+
```
liornoy, line 116: I made a similar function in my PR but it's not a class method, but a model function for a question that returns QuerySet of its answers filtered by date or votes. so maybe this function will be unnecessary

home/tests.py
idokk1: Thanks, removed it

home/models.py
idokk1: I saw your function, it returns all the answers to a particular question.
My test here is to fetch all the answers from DB and see if it matches what we expect.

idokk1:
@ifireball can you review these tests, 
I will make another PR with tests on view when I will make the add answer feature.

ifireball:
I'm looking at the PR description and I'm not understanding the point of the Pr or what it is supposed to contain any more. Its it just fr adding tests?

idokk1:
> I'm looking at the PR description and I'm not understanding the point of the Pr or what it is supposed to contain any more. Its it just fr adding tests?

this PR includes DB tests on adding new answer and adding create new profile.
PR #87 includes the add answer feature with the editor and view and tests for that (test the post method).

I made this PR before Edward's lecture, so the feature wasn't ready then.

ifireball:
I'm still not understanding. Is this PR adding tests for code you already merged of for code that you have in another PR? 

Is there PR no non-test code in this PR at all?

idokk1:
> I'm still not understanding. Is this PR adding tests for code you already merged of for code that you have in another PR?
> 
> Is there PR no non-test code in this PR at all?

this PR is tests: 
1. Add profile (not related to any other PR)
2. DB tests to adding answer - related to #87  PR. 

 in #87 PR there are tests for the view and the post method for adding answer.

ifireball:
It does not make sense to have a PR just for tests (Except to the case where you are closing a technical debt of adding tests to code that you merged in the past without tests).

If this PR is adding tests for functionality added in another PR - put the tests in that other PR. 

If this PR is also introducing functionality elements, please clearly state those elements in the PR description and make sure to onlu include tests for those elements.

idokk1:
> It does not make sense to have a PR just for tests (Except to the case where you are closing a technical debt of adding tests to code that you merged in the past without tests).
> 
> If this PR is adding tests for functionality added in another PR - put the tests in that other PR.
> 
> If this PR is also introducing functionality elements, please clearly state those elements in the PR description and make sure to onlu include tests for those elements.

OK, so I transferred all the tests to #87 PR , and I will close this PR.

PR #65, opened 2021-04-15
Add a basic new question page and tests for that

### LINKED ISSUE #60: Add a basic 'Add a question' page using model form
Add a basic 'Add a question' page.
In this issue, the page will not include style, tags, and special content editor - only the basic action of posting a question using model form. 
files that need to be modified :
- home/urls.py
- home/views.py
 
new files :
- home/forms.py
- home/templates/home/questions/new_question.html

avivz450:
Used a model form when made this page. 
This is a basic 'add a question' page that stores any submitted question in the database.
This PR **does not** contain content editor, styling, JS elements, redirections, success/failure messages of inserting a question, and tags.

- Edited and added files to add the "insert question" feature
- Altered Question model and added migration file for that
- Added tests for adding a new question

fix #60 #79


home/templates/home/questions/new_question.html
```diff
@@ -0,0 +1,28 @@
...
+<html lang="en">
+
+<head>
+    <meta charset="UTF-8">
+    <meta http-equiv="X-UA-Compatible" content="IE=edge">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <title>Document</title>
+</head>
+
+<body class="landing-bg">
```
liornoy, line 14: Since you extend "home/base.html" you don't need to write the html, head, and body tags
they all written already in base.html.
so the first tag should be div class=newfrm
Also for adding a title look at the comment I wrote on views.py

home/views.py
```diff
@@ -7,3 +9,17 @@ def about(request):
...
+    if request.method == 'POST':
+        questForm = QuestionForm(request.POST)
+        if questForm.is_valid():
+            questForm = questForm.save(commit=False)
+            for currProfile in Profile.objects.all():
+                if (currProfile.user == request.user):
+                    questForm.profile = currProfile
+                    break
+            questForm.save()
+    return render(request, 'home/questions/new_question.html', {'form': form})
```
liornoy, line 25: you can add here next to 'form':form another variable - 'title': '<title-name>'
in order to edit the title of the new question page that will be shown on top of the tab in the browser

home/templates/home/questions/new_question.html
avivz450: I agree with you about the html and head, but as soon as I delete the body the background color we set will also be deleted. Do you have a way for the background color (defined by the "landing-bg" class) to remain while I delete the body?
Also, logically, I add the feature of adding the question to the body of the page. Do you still think it's right to delete the body completely from new_question.html?

home/views.py
avivz450: Ok, I added 'title': 'New Question'

home/views.py
```diff
@@ -7,3 +9,17 @@ def about(request):
...
+    if request.method == 'POST':
+        questForm = QuestionForm(request.POST)
+        if questForm.is_valid():
+            questForm = questForm.save(commit=False)
+            for currProfile in Profile.objects.all():
+                if (currProfile.user == request.user):
+                    questForm.profile = currProfile
+                    break
+            questForm.save()
+    return render(request, 'home/questions/new_question.html', {'form': form, 'title': 'New Question'})
```
idokk1, line 25: I think after submitting the form you need to go back to Lior's page (question-detail) of the new question.
you can change it after Lior's PR will be merged.

avivz450: You're right, but I thought of making the links in a different PR (with more things). 
Anyway, if Lior's PR will approve before this, I'll do it here.

home/templates/home/questions/new_question.html
liornoy: add  class="main-bg" to the body tag in base.html and it will take care of the background-color
the feature of adding the question will still be in the body of the page because you extend the base.html
where the " {% block content %}{% endblock %} " lays inside a body tag.

avivz450: I saw that there is no 'body' tag in base.html.
As far as I know, you should update this file now and also later.
I thought about deleting the body tag from new_question.html and after you finish updating the base.html the updates will be done automatically in new_question.html.
What do you think?

liornoy: sounds good

liornoy: Aviv, I was wrong, my apologies. you should add a body tag in your HTML and give it a background color.
In my new PR #71 "Fix html.base" I removed the color from the body, so this CSS won't apply to all pages and everyone will have control over the background color in their feature.
Moreover, writing a body tag inside your template - which extends a template that already has a body tad, won't make a corrupt structure or anything like that. it will merge the body tags into one - at least as I saw during my work.

avivz450: Ok, changed

ifireball (changes requested):
Please add tests (You might need to restructure the code a bit to do that...)

avivz450:
> Please add tests (You might need to restructure the code a bit to do that...)



> Please add tests (You might need to restructure the code a bit to do that...)

Added tests for that

avivz450:
> Please add tests (You might need to restructure the code a bit to do that...)

Added 👍

ifireball:
Please rebase and clean up the commit history.

avivz450:
> Please rebase and clean up the commit history.

Done

avivz450:
Rebased and ready for review @ifireball @AviadP

home/tests.py
```diff
@@ -96,3 +99,100 @@ def question_tag_test_data(question_test_data):
...
+    # User chose negative book number
+    (10, 1, "Question in Math", 'How much is it 1+1?',
+     datetime(2022, 4, 7, 12, 53, 29, 4, tzinfo=pytz.UTC), 1, 4, Grade.GRADE7, 1, 99990, False),
+    # User entered to high book number
+])
+# checks if the question was not inserted
+@pytest.mark.django_db
+def test_add_invalid_question(invalid_Input):
+    invalid = False
+
```
ifireball, line 162: To test for exceptions, please use `pytest.raises`. If different inputs are expected to raise different types of exceptions, you can pass the expected exception type is part of the "parametrize" structure.

home/tests.py
```diff
@@ -96,3 +99,100 @@ def question_tag_test_data(question_test_data):
     new_pair.tag = test_tag
     new_pair.save()
     return test_tag
+
```
ifireball, line 102: As I've suggested elsewhere, please group your tests in classes because you're going to have many of them in this file

home/views.py
```diff
@@ -7,3 +9,17 @@ def about(request):
 
 def landingpage(request):
     return render(request, 'home/landingpage.html')
+
+
+def new_question(request):
```
ifireball, line 14: Where are the tests for this view?

ifireball (changes requested):

home/views.py
```diff
@@ -7,3 +9,17 @@ def about(request):
...
+    form = QuestionForm
+    if request.method == 'POST':
+        questForm = QuestionForm(request.POST)
+        if questForm.is_valid():
+            questForm = questForm.save(commit=False)
+            for currProfile in Profile.objects.all():
+                if (currProfile.user == request.user):
+                    questForm.profile = currProfile
```
Idokah, line 22: I suggest you to change the for loop to this line :

questForm.profile = Profile.objects.filter(user=request.user).first()

home/views.py
```diff
@@ -7,3 +9,17 @@ def about(request):
...
+                    break
```
idokk1, line 23: I suggest you to replace the for loop to this line: 

questForm.profile=Profile.objects.filter(user=request.user).first()

avivz450: Thank you, I will update

ifireball: This is a good candidate for a class method, but since you've defined the link between the models I wouldn't be surprised if the ORM already generates something like this for you.

Also since you're probably almost always going to need the profile information alongside the user information in your app you may want to create a middleware component that lodes it automaticallt and stores in request just like the user information is stored.

avivz450: So instead of making a class method, you want me to create a middleware component so I can call to request.profile at views file?

ifireball: yep, seems like it would be a useful think to have througout your application if you'll have more then one view that needs information from the profile

avivz450: Got it, ill do it.

avivz450:
ready for another review @ifireball

home/tests.py
```diff
@@ -193,3 +197,105 @@ def test_response_context(self, response):
...
+            assert question.sub_subject is None
+
+    class TestViews:
+        @pytest.mark.django_db
+        def test_uses_question_form(self, client):
+            response = client.get('/explore/new_question')
+
+            assert response.status_code == 200
+            assert isinstance(response.context['form'], type(QuestionForm))
```
ifireball, line 294: missing a test for the various flows of posting data - when proper data is posted, and when bad data is posted...

ifireball (changes requested):
missing test coverage

home/tests.py
avivz450: I made these tests at TestDatabaseInsertions class, why do I have to do it also in TestViews class? what is the difference?

avivz450:
> missing test coverage

Added and ready for another review @ifireball @AviadP

ifireball:
please resolve conflicts

avivz450:
> please resolve conflicts

Resolved

home/templates/home/questions/new_question.html
```diff
@@ -0,0 +1,16 @@
+{% extends "home/base.html" %}
+{% load static %}
+{% block content %}
+
+<body class="landing-bg">
```
liornoy, line 5: About the class landing-bg. it's a very light color used for the landing page.
I think you should change it to the 'main-bg' class so it will match the design in the Wireframe, and have
identical bg color to the display question page and the tags page.
Also, I think giving the body tag this class should be sufficient and you don't need to write it also in the div below

avivz450: Changed

home/tests.py
```diff
@@ -205,3 +208,149 @@ def tags_response(self, client):
...
+            with pytest.raises(exception):
+                assert Question(*invalid_Input).clean_fields()
+
+        # checks default value in is_edited
+        @pytest.mark.django_db
+        def test_default_is_edited(self, question):
+
+            assert question.is_edited is False
+
+        # checks default value in book_page once the user didn't insert a book page
```
ifireball, line 281: Instead of writing such comments, why don't you jut call the function: `test_default_book_page_is_none` ?

Comments can lie and go out of sync with the code quickly, if the code can document itself, do that instead of using comments.

home/tests.py
```diff
@@ -205,3 +208,149 @@ def tags_response(self, client):
...
+              'subject': 1, 'grade': Grade.GRADE7, 'book_page': 23}),
+            # User entered all the required fields and a valid book page
+        ])
+        @pytest.mark.django_db
+        def test_post_valid_question_with_client(self, client, valid_data):
+            latestInsertedQuestionID = Question.objects.latest('id').id
+            client.login(username='Lior', password='LiorLior')
+            client.post('/explore/new_question', data=valid_data)
+
+            assert Question.objects.latest('id').id != latestInsertedQuestionID
```
ifireball, line 320: this check is insufficient, how do you know that data you;ve posted actually made it to the DB?

home/tests.py
```diff
@@ -205,3 +208,149 @@ def tags_response(self, client):
...
+            ({'title': "Question in Math", 'content': 'How much is it 1+1?', 'subject': 1, 'book_page': 43343}),
+            # User entered all the required fields and to high book page
+        ])
+        @pytest.mark.django_db
+        def test_post_invalid_question_with_client(self, client, invalid_data):
+            latestInsertedQuestionID = Question.objects.latest('id').id
+            client.login(username='Lior', password='LiorLior')
+            client.post('/explore/new_question', data=invalid_data)
+
+            assert Question.objects.latest('id').id == latestInsertedQuestionID
```
ifireball, line 342: I don't see any checking about the output the user gets? Is an error message being displayed? How?

home/tests.py
```diff
@@ -205,3 +208,149 @@ def tags_response(self, client):
...
+
+        @pytest.mark.django_db
+        def test_uses_question_form(self, client):
+            response = client.get('/explore/new_question')
+
+            assert response.status_code == 200
+            assert isinstance(response.context['form'], type(QuestionForm))
+
+        @pytest.mark.django_db
+        def test_renders_add_question_template(self, client):
```
ifireball, line 352: it looks like this test can be merged with the one above it, I don't think there is a need to test the context and the template in separate tests.

home/profile_middleware.py
```diff
@@ -0,0 +1,15 @@
+from .models import Profile
+
+
+class ProfileMiddleware:
```
ifireball, line 4: No seeing any tests for this

home/tests.py
avivz450: > it looks like this test can be merged with the one above it, I don't think there is a need to test the context and the template in separate tests.

Ok, I will merge them

avivz450:
Hey @ifireball, I made some of the changes you requests.
My goal in this PR is to make the most basic feature of entering a question, and then make small additions to it in additional PRs after it will be approved.
I did not do some of the tests you requested because I did not implement these things in this PR.
Regarding the Middleware - it was difficult to simulate situations of a connected user, a disconnected user, and Requests.
Therefore, I "imitated" certain situations to do tests for it. If it's wrong, can you please explain to me how to do tests for it?

Pipfile
```diff
@@ -11,6 +11,7 @@ pytest-django = "*"
 [packages]
 django = "*"
 pytest = "*"
+mock = "*"
```
ifireball, line 14: No need for this, unitest.mock is included in the Python3 standard library.

home/tests.py
```diff
@@ -7,6 +8,9 @@
 from pytest_django.asserts import assertTemplateUsed
 import pytest
 import pytz
+from django.core.exceptions import ValidationError
+from home.profile_middleware import ProfileMiddleware
+from mock import Mock
```
ifireball, line 13: use unitest.mock from the standard library.

home/tests.py
```diff
@@ -208,3 +212,171 @@ def tags_response(self, client):
...
+            assertTemplateUsed(response, 'home/questions/new_question.html')
+
+
+class TestProfileMiddleware():
+    class LoggedOutUserMock:
+        is_authenticated = False
+
+    @pytest.mark.django_db
+    def test_process_request_with_logged_in_user(self):
+        LoggedInUser = User.objects.latest('id')
```
ifireball, line 366: is this passing flake8? In Python, variable names are usually snake_case.

home/tests.py
```diff
@@ -208,3 +212,171 @@ def tags_response(self, client):
...
+        @pytest.mark.django_db
+        def test_question_form_and_template_displayed(self, client):
+            response = client.get('/explore/new_question')
+
+            assert response.status_code == 200
+            assert isinstance(response.context['form'], type(QuestionForm))
+            assertTemplateUsed(response, 'home/questions/new_question.html')
+
+
+class TestProfileMiddleware():
```
ifireball, line 360: This is a a somewhat fragile test, since it rlies too heavily on the internal implementation of the middleware. It is also not verifying the the correct functionality is achieved from the point of view of a view function.

The way I would test it would be:
1. Use `monkeypatch` to replace a view function with a mockup:
2. Perform requests using the `client` fixture and its [`force_login` and `logout` methods](https://docs.djangoproject.com/en/3.2/topics/testing/tools/#django.test.Client.force_login)
3. Inspect the `request` structure that was passed to the view mockup to assert the existance and the contents of the `profile` attribute.

Given that the test does partially work, you may decide not to implement this at this point.

avivz450: At the moment I prefer to leave the implementation as it is, as long as it is acceptable enough, and later @liornoy and I together will improve the test in the way you mentioned.
I opened an issue for that.

avivz450:
Ready for another review @ifireball

ifireball (approved):

PR #67, opened 2021-04-16
Add a js script to question page

### LINKED ISSUE #66: Add JS file to 'new_question' page
Need to do it to display certain fields within a Django form based on the data of another selected field within that same form.
For an example :
If the User chose 'Math' in the subject field, he will only see all the sub-subjects that related to math (such as 'Geomtry' and more...)

avivz450:
- Added new_question.js to new question page
- Loaded the JS file into new_question.html

fix #66

### DIFF
--- a/home/static/home/js/new_question.js
+++ b/home/static/home/js/new_question.js
@@ -1,0 +1,153 @@
+function disableAllSubSubject() {
+    ID_Sub_Subject.value = ""
+    for (let i = 0; i < ID_Sub_Subject.length; i++) {
+        if (i !== 0) {
+            ID_Sub_Subject[i].style.display = 'none';
+        }
+    }
+}
+
+function disableAllBooks() {
+    ID_Book.value = "";
+    ID_Book_Page.value = "";
+    ID_Book_Page.disabled = true;
+    for (let i = 0; i < ID_Book.length; i++) {
+        if (i !== 0) {
+            ID_Book[i].style.display = 'none';
+        }
+    }
+}
+
+function showBibleSubSubjectsAndBooks() {
+    ID_Sub_Subject[1].style.display = "";
+    ID_Sub_Subject[2].style.display = "";
+    ID_Sub_Subject[3].style.display = "";
+    ID_Book[2].style.display = "";
+    ID_Book[3].style.display = "";
+    ID_Book[4].style.display = "";
+}
+
+function showBiologySubSubjectsAndBooks() {
+    ID_Sub_Subject[4].style.display = "";
+    ID_Sub_Subject[5].style.display = "";
+    ID_Sub_Subject[6].style.display = "";
+    ID_Book[1].style.display = "";
+}
+
+function showChemistrySubSubjectsAndBooks() {
+    ID_Sub_Subject[7].style.display = "";
+    ID_Sub_Subject[9].style.display = "";
+}
+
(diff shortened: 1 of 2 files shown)


danitLevi:
I think you can close this PR because those commits are also in PR #68

avivz450:
> I think you can close this PR because those commits are also in PR #68

I prefer not to close this PR so that each update of the question insertion feature will be in a separate PR.
The reason you see a lot of commits in this PR and also in PR #68 is because PR #65 is not yet merged.
After PR #65 will be merged and PR #67 and PR # 68 will be rebased, it will be more comfortable to see the changes made in each PR, and some of the commits you see now will not be there.
Accordingly, the status of these PRs (#67 #68) will also change and will not be "in progress".

ifireball (changes requested):
How would you test these 150 lines of code?

avivz450:
> How would you test these 150 lines of code?

Trying to figure it out. I'm not sure how to do a test for this addition.
I do not find a way to check what options the user has when he selects a particular option in the form in other fields there.
I tried to combine tests with [QUnit](https://qunitjs.com/) according to the guide [here](https://docs.djangoproject.com/en/dev/internals/contributing/writing-code/javascript/) but it did not help so much.
Do you have a recommendation/guidance?

ifireball:
> Do you have a recommendation/guidance?

My recommendation is to be more creative on the server side and avoid JavaScript. 

When a project decides to adopt a new language, it needs to be done carefully while making sure all the needed tooling is available to avoid degrading the quality of the dev process. In other words, if you really want to write JS on your own, you need to setup a full JS tool-chain, and probably also adopt a more collaboration/testing friendly language that compiles into JS such as TypeScript or Kotlin. But depending on the size of the team that may not be a smart move (You essentially split your team into writing two appllcations - a browser one and a server one).

If you just want to add some inter-activeness you can try things like [sock-puppet](https://github.com/jonathan-s/django-sockpuppet) or [HTMX](https://htmx.org/) which let you avoid writing JS for the most part.

Another option is to use something like [Brython](https://brython.info/) to run Python in the Browser.

But frankly, I wouldn't do any of those at this stage of the development, given the existing deadlines, and just focus on getting the needed functionality without any extra client-side bells and whistles.

avivz450:
> > Do you have a recommendation/guidance?
> 
> My recommendation is to be more creative on the server side and avoid JavaScript.
> 
> When a project decides to adopt a new language, it needs to be done carefully while making sure all the needed tooling is available to avoid degrading the quality of the dev process. In other words, if you really want to write JS on your own, you need to setup a full JS tool-chain, and probably also adopt a more collaboration/testing friendly language that compiles into JS such as TypeScript or Kotlin. But depending on the size of the team that may not be a smart move (You essentially split your team into writing two appllcations - a browser one and a server one).
> 
> If you just want to add some inter-activeness you can try things like [sock-puppet](https://github.com/jonathan-s/django-sockpuppet) or [HTMX](https://htmx.org/) which let you avoid writing JS for the most part.
> 
> Another option is to use something like [Brython](https://brython.info/) to run Python in the Browser.
> 
> But frankly, I wouldn't do any of those at this stage of the development, given the existing deadlines, and just focus on getting the needed functionality without any extra client-side bells and whistles.

Ok, thanks for the detailed answer. I will close the PR

PR #70, opened 2021-04-19
Created Many-To-Many Relationship Tag & Question

rebeccaTubman:
Added to class 'Question':
- field:
  tags = models.ManyToManyField('Tag', through='Question_Tag')

Added to class 'Tag':
- field:
  questions = models.ManyToManyField('Questions', through='Question_Tag')
- model function:
  - add_tags - the input is a set of strings that represent tag names.
    The function creates and saves to the DB all the tags that do not
    exist yet in the DB.
  - tags_feed - the function returns a set of tags ordered by
    'tag_name' and filtered by the input or the default value.

Added to class 'Question_Tag':
- model function:
  - add_quetion_tags - the input is a set of strings that represent
    tag names and integer that represents question_id.
    The function creates a relation between all the tags and the
    question.

Modified: home/models.py
New file: home/migrations/0013_auto_20210414_1712.py

Notes:
* The added fields create virtual sets that will allow easy access to
  the related tags of a particular question and questions of a
  particular tag with: object_name.field_name.values().


danitLevi:
LGTM

liornoy:
Looks good

home/models.py
```diff
@@ -104,11 +105,28 @@ def __str__(self):
...
-    tag_name = models.CharField(max_length=50)
+    tag_name = models.CharField(max_length=20)
+    questions = models.ManyToManyField('Question', through='Question_Tag')
 
     def __str__(self):
         return self.tag_name
 
+    # Input: array of strings that represent name of tags.
+    # The function will go through the array and will add them to the DB if they don't exist yet.
+    @classmethod
```
ifireball, line 116: not sure what this funciton would be useful for - if its to let a user type-in a list of tags to be added to a question, then it would probably need to either return the tag objects, get a question object as input or be a method of Question.

I suggest you add model methods only when you have a very clear idea in your mind about where and how they would be used.

home/models.py
```diff
@@ -122,3 +140,16 @@ class Meta:
 
     def __str__(self):
         return f"{self.question}, Tag : {self.tag}"
+
+    # Input: array of strings that represent name of tags and a question.
+    # The function will go through the array and will add them as a pair - question and tag.
```
ifireball, line 145: Why not add this is a regular method in Question? I think it would make more sense then passing `question_id` in here and having to query the DB for it...

home/models.py
```diff
@@ -122,3 +140,16 @@ class Meta:
...
+    @classmethod
+    def add_quetion_tags(cls, tags, question_id):
+        question = Question.objects.get(pk=question_id)
+        for tag_name in tags:
+            tag = Tag.objects.get(tag_name=tag_name)
```
ifireball, line 150: you are running a separate DB query for each and every tag, think if you can somehow optimize this to reduce the amount of queries.

ifireball (changes requested):

ifireball (changes requested):
Where are the tests? Either add tests to the new logic functions or drop them.

rebeccaTubman:
> Where are the tests? Either add tests to the new logic functions or drop them.

I'm working on the Tags UI. Is it possible to add all the tests there?

ifireball:
> > Where are the tests? Either add tests to the new logic functions or drop them.
> 
> I'm working on the Tags UI. Is it possible to add all the tests there?

Possible, but not advisable. The point of adding logic funcitons to the data model is to enable testing the logic in a way that is separate from the UI concerns. Please take advantage of that.

rebeccaTubman:
I've added tests for the new logic functions:)

rebeccaTubman:
I think this PR is ready for a re-review @ifireball @AviadP

home/tests.py
```diff
@@ -1,3 +1,79 @@
...
+
+    # check if (test_question, test_tag_2) in Question_Tag
+    assert Question_Tag.objects.filter(question=test_question, tag=test_tag).exists()
+
+    # check test_tag_3 wasn't added again
+    test_question.add_tags_to_question(['test_tag_3'])
+    assert test_question.tags.values().count() == 3
+
+    # delete test data
+    delete_test_data(test_question.pk, test_tags)
```
ifireball, line 34: you don't need this - all the data that you introduce gets deleted automatically between tests since you use the django_db decorator.

home/tests.py
```diff
@@ -1,3 +1,79 @@
...
+
+    # check if the function returns empty set when tags
+    assert Tag.tags_feed('testtesttesttest').count() == 0
+
+    # check if the function is up to date with DB after deletion
+    Tag.objects.filter(tag_name='test_tag_1').delete()
+    assert Tag.tags_feed().count() == 4
+
+
+def create_test_data():
```
ifireball, line 58: this kind of function is essentially a fixture, please write it as such.

home/tests.py
```diff
@@ -1,3 +1,79 @@
...
+
 
 def test_first():
     assert True
+
+
+@pytest.mark.django_db
+def test_add_tags_to_question():
+    # create test data
+    test_question, test_tags = create_test_data()
```
ifireball, line 14: I would leave the definition of "tags" in the body of the test function so its easier to find out what you are passing to the function you are testing.

home/tests.py
```diff
@@ -1,3 +1,79 @@
...
+    # check test_tag_3 wasn't added again
+    test_question.add_tags_to_question(['test_tag_3'])
+    assert test_question.tags.values().count() == 3
+
+    # delete test data
+    delete_test_data(test_question.pk, test_tags)
+
+
+@pytest.mark.django_db
+def test_tags_feed():
```
ifireball, line 38: When you write a block of code and you find yourself adding a comment on top of it to explain what it does - its usually a hint that you should be putting the code in its own function.

Please break this test apart into separate tests so that the function name of each test tells you what it is testing - it will also make finding out what exactly broke when your tests fail much easier.

rebeccaTubman: I managed to break the function only in to 3 tests, because a few checks depend on the test tag I create in the function.

rebeccaTubman:
@ifireball , I followed your instructions and committed the changes.

ifireball:
> @ifireball , I followed your instructions and committed the changes.

I still see blocks of code with "header" comments in your tests - for example  `test_tags_feed_with_test_tag` has 3 assertions in it that seem unrelated - those probably should be different tests.

If you feel like the code you wrote is not clear enough and you need to write a comment to explain it - its a good indicator that you perhaps need to refactor the code so its clearer.  For example, by selecting better method names.

rebeccaTubman:
> > @ifireball , I followed your instructions and committed the changes.
> 
> I still see blocks of code with "header" comments in your tests - for example `test_tags_feed_with_test_tag` has 3 assertions in it that seem unrelated - those probably should be different tests.
> 
> If you feel like the code you wrote is not clear enough and you need to write a comment to explain it - its a good indicator that you perhaps need to refactor the code so its clearer. For example, by selecting better method names.

`test_tags_feed_with_test_tag` has been split so each will have only 1 assert.
I'm having trouble splitting `test_add_tags_to_question` - all the asserts are connected logically, they all check if the many-to-many logic and the function logic work. Do you have any suggestion?

ifireball:
> > > @ifireball , I followed your instructions and committed the changes.
> > 
> > 
> > I still see blocks of code with "header" comments in your tests - for example `test_tags_feed_with_test_tag` has 3 assertions in it that seem unrelated - those probably should be different tests.
> > If you feel like the code you wrote is not clear enough and you need to write a comment to explain it - its a good indicator that you perhaps need to refactor the code so its clearer. For example, by selecting better method names.
> 
> `test_tags_feed_with_test_tag` has been split so each will have only 1 assert.
> I'm having trouble splitting `test_add_tags_to_question` - all the asserts are connected logically, they all check if the many-to-many logic and the function logic work. Do you have any suggestion?

When I look at the test, only the first assert seems to test the functionality of the `add_tags_to_question` the rest seem to test various aspects of having tags linked to questions, so perhaps have such links in your test data or create them in a separate fixture to test those things in their own tests. 

I still see the comment in your code - please remove them, if you still think you need them it means you probably need to give better names to the test functions.

rebeccaTubman:
@ifireball thank you for the suggestion:)
All of the test include only 1 assert each.
I deleted the comments and won't add comments from now on. 
@AviadP

ifireball (approved):

Issue #48, opened 2021-04-14
Add "Explore page"


danitLevi:
Add "Explore page" - a page that displays all the questions in the app.

PR #74, opened 2021-04-22
Creating "ask new question" button in "explore page"

### LINKED ISSUE #51: Update links in Explore-page
- Add "add question" button that leads to "add question page" (should be synchronized with @avivz450 )

- Click on question in explore page leads to the selected question page (should be synchronized  with @liornoy )

danitLevi:
- Creating "ask new question" button in explore page . Clicking on it leads to "new question" page
- Testing "ask new qiestion" btn :
  - Asserting that the response content contains the url of "new question"   page

fix #51

### DIFF
--- a/home/templates/home/explore.html
+++ b/home/templates/home/explore.html
@@ -2,19 +2,24 @@
 {% load static %}
 {% block content %}
 
-<script src="//ajax.googleapis.com/ajax/libs/jquery/1.9.1/jquery.min.js"></script>
-<script type="text/javascript" src="{% static 'home/js/explore_page_js.js' %}"></script>
 <link rel="stylesheet" type="text/css" href="{% static 'home/CSS/explore-page.css' %}">
 <body class="main-bg">
 
 <form method="GET" class="container">
 <div class="m-5" >
 
+    <div class="row mb-4">
+        <div class="col-2">
+            <a href="{% url 'new_question' %}" class="btn shlifim-btn" role="button">Ask new question</a>
+        </div>
+
+    </div>
+
 	{% for question in questions %}
 	<div class=" question-display p-2 row m-0 border">
 	    <img class="subject-img" src="{% get_static_prefix %}home\subjects-images\{{question.subject|lower}}.png">
             <div class="col-3">
-                <a style="color:#6FB5EE" href="{% url 'question-detail' question.id %}">{{question.title }}</a>
+                <a style="color:#6FB5EE" href="{% url 'question-detail' question.id %}"><b>{{question.title }}</b></a>
             </div>
             <div class="col">
                 <span>{{ question.subject }} ,{{ question.sub_subject }}</span>
--- a/home/static/home/CSS/base_style.css
+++ b/home/static/home/CSS/base_style.css
@@ -18,3 +18,7 @@ body {
 	color: #000000;
 	padding-top: 20px;
 }
+
+.shlifim-btn{
+  background-color: #A9D4D9;
+}
--- a/home/static/home/CSS/explore-page.css
+++ b/home/static/home/CSS/explore-page.css
@@ -10,10 +10,6 @@
   background-color: #f3ebe6;
 }
 
-.shlifim-btn{
-  background-color: #A9D4D9;
-}
-
 .landing-bg{
 	background-color: #f2dccb;
 }
--- a/home/tests/test_explore_page.py
+++ b/home/tests/test_explore_page.py
@@ -34,3 +34,8 @@ class TestExplorePage:
 
             questions = Question.objects.all()
             assert set(explore_page_response.context['questions']) == set(questions)
+
+        @pytest.mark.django_db
+        def test_new_question_btn_in_explore_page(self, explore_page_response):
+            char_content = explore_page_response.content.decode(explore_page_response.charset)
+            assert '<a href="%s"' % reverse("new_question") in char_content


ifireball (changes requested):
This is a huge PR - please break it apart. Usally when you see you're getting a single PR that fixes more then one issue - its a good indicator you've included too much stuff in the same PR.

avivz450 (commented):
I think you should edit the models.py and add min and max value to the book_page field in the "Question" model. That way the user will not be able to filter a book that his page is negative, or too high.
You can do it by adding `from django.core.validators import MinValueValidator, MaxValueValidator` at the beginning of models.py and by adding ` validators=[MinValueValidator(1), MaxValueValidator(999)]` inside models.IntegerField in book_page field.

home/models.py
```diff
@@ -88,6 +88,9 @@ class Meta:
     def __str__(self):
         return f"Question #{self.id} : {self.title} ( {self.publish_date.strftime('%d/%m/%Y %H:%M')} )"
 
+    def get_answers_num(self):
+        return self.answer_set.count()
+
```
liornoy, line 93: What is answer_set? we don't have that field in Question model

home/templates/home/explore.html
```diff
@@ -0,0 +1,75 @@
...
+.landing-bg{
+	background-color: #f2dccb;
+}
+
+.collapsed-bg
+{
+    background-color: #f3ebe6;
+}
+
+</style>
```
liornoy, line 31: I wrote all the style in the HTML also, but Barak suggest that we move it do a different CSS file in the Static folder

.idea/.gitignore
```diff
@@ -0,0 +1,3 @@
+# Default ignored files
```
liornoy, line 1: What are these files in .idea folder are for?

liornoy (commented):
I saw some things to comment about but I think in order to really understand the changes, you should squash all the commits so the HTML file will be complete

.idea/.gitignore
danitLevi: @liornoy 
I usually use PyCharm and it automatically creates "idea" folder that contains settings for the project. I accidentally added it to the PR. I will remove it and add it to the git ignore so it won't happen again.

home/models.py
danitLevi: Yes, We don't have answer_set field in Question model .

The _set suffix is a reverse lookup class variable in Django.
You can read more about it in the following link https://docs.djangoproject.com/en/3.2/topics/db/examples/many_to_one/

You're welcome to ask me if you have any further questions regarding this issue .

home/templates/home/explore.html
danitLevi: I moved it to different css file as requested

danitLevi:
> I think you should edit the models.py and add min and max value to the book_page field in the "Question" model. That way the user will not be able to filter a book that his page is negative, or too high.
> You can do it by adding `from django.core.validators import MinValueValidator, MaxValueValidator` at the beginning of models.py and by adding ` validators=[MinValueValidator(1), MaxValueValidator(999)]` inside models.IntegerField in book_page field.


@avivz450 
I add the min and max value to the book_page field in the filter form. 
(I didn't change the question model as you suggested because it didn't work for filtering, but I change the filter set I created so the user will not be able to filter book page by negative or too high value)

danitLevi:
> This is a huge PR - please break it apart. Usally when you see you're getting a single PR that fixes more then one issue - its a good indicator you've included too much stuff in the same PR.

Hi @ifireball 

At first, I installed Pytest in this PR and added "pytest.ini" and ".github/workflows/pytest.yml" files. I removed them now in order to prevent conflicts (I saw Ido already added them in the PR he created )

Besides, I usually use PyCharm and it automatically creates an "idea" folder that contains settings for the project. It seems that initially I accidentally added it to the PR, and maybe it causes the PR to look huge.

Moreover, I moved the tests I created to the "explore page" into different PR. 

Is this PR OK now? Or is it still too big?
I know this PR handling multiple issues but all this issues are related to the same page (explore page ) and each commit is based on the previous commit . Should I separate them into separated PRs?

In addition, I have a question. Before changing this PR I created PR #75 on top of it (because I needed the URL of explore page in this PR). Therefore the commits from this PR also appear in PR #75, but now the commits in PR #75 didn't change automatically according to the changes I made.  Can it cause problems after merging this PR ? Should do anything else in PR #75 in this case?

danitLevi:
> I saw some things to comment about but I think in order to really understand the changes, you should squash all the commits so the HTML file will be complete

@liornoy  You can see the final files by clicking on the "fiels changed"

ifireball:
@danitLevi about the size of the PR - the issues you've created actually represent a good division of topics - each one of them should be its own PR, with its own tests.

Doing it like that means that once you wrote a PR for issue #48 you can do what you needed to do in PR #75 without having to implement the rest.

This might mean you may not finish implementing everything you have here for the demo but that is fine, its a good lesson about how much we think we need to implement v.s. how much we actually need.

home/models.py
liornoy: Cool didn't know it. Thanks!!

danitLevi:
> @danitLevi about the size of the PR - the issues you've created actually represent a good division of topics - each one of them should be its own PR, with its own tests.
> 
> Doing it like that means that once you wrote a PR for issue #48 you can do what you needed to do in PR #75 without having to implement the rest.
> 
> This might mean you may not finish implementing everything you have here for the demo but that is fine, its a good lesson about how much we think we need to implement v.s. how much we actually need.

@ifireball 
I split my PR into several PRs as you requested.
For each PR I added the commits it depends on. Did I create them ok?
You can review them in the following order :
PR #80  --- >  PR #75  --- >  PR #81 --- >  PR #82 --- >  PR #83 --- >  PR #74

danitLevi:
@ifireball  @AviadP

This PR is ready for review

rebeccaTubman:
Looking good

ifireball:
Team: can you please request and do review properly and not just comment?

ifireball (approved):
LGTM but missing team reviews

rebeccaTubman:
> Looking good

I think this can be merged.

ifireball:
@rebeccaTubman why are you not posting a review approval?

home/templates/home/explore.html
```diff
@@ -10,6 +10,13 @@
 <form method="GET" class="container">
 <div class="m-5" >
 
+    <div class="row mb-4">
```
rebeccaTubman, line 13: A great PR with a small feature with test coverage.

rebeccaTubman (approved):

liornoy (approved):

avivz450 (approved):

Issue #59, opened 2021-04-14
Add 'Tags' Page


rebeccaTubman:
The page will display all the tags and will allow to search for a specific tag.
The design for the Tags page will be similar to Users page from the pages_design.md in documentation.

![image](https://user-images.githubusercontent.com/50530087/113506873-f036a480-954f-11eb-872a-3e829d407573.png)

PR #85, opened 2021-04-29
Adding Tags Page view & tests

rebeccaTubman:
- Display tags feature with view test.
- Added all the many-to-many in to one class and the tags view test in
  to another class.

modified:   home/tests.py
                  home/urls.py
                  home/views.py

new file:   home/templates/home/tags.html

fix: #59

### DIFF
--- a/home/templates/home/tags.html
+++ b/home/templates/home/tags.html
@@ -1,0 +1,34 @@
+{% extends "home/base.html" %}
+{% load static %}
+{% block content %}
+<link rel="stylesheet" type="text/css" href="{% static 'home/CSS/tags_style.css' %}">
+<html>
+    <body>
+        <main role="main" class="container">
+        <div class="row">
+            <div class="col-md-12">
+            <br><br>
+            <center><h1>Tags</h1></center>
+            <br><br>
+            <div class="tags-grid">
+                <form class="d-flex">
+                    <input name="q" class="form-control me-2" type="search" placeholder="Search" aria-label="Search">
+                    <button class="btn btn-light" type="submit" ng-click="ctrl.clearSearch()">Search</button>
+                </form>
+                <br><br>
+                <div class="wrapper">
+                    {% for tag in tags %}
+                    <div class="card border-secondary mb-3">
+                        <div class="card-body">
+                            <p class="card-text">{{tag.tag_name}}</p>
+                        </div>
+                    </div>
+                    {% endfor %}
+                </div>
+            </div>
+            </div>
+        </div>
+    </body>
+   
+</html>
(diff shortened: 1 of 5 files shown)


home/templates/home/tags.html
```diff
@@ -0,0 +1,60 @@
...
+            margin-top: 20px;
+        }
+        .item-in-wrapper {
+            cursor: pointer;
+            position: relative;
+            height: 200px;
+            width: 180px;
+            text-align: center;
+        }
+    </style>
```
liornoy, line 30: Barak recommended having all the styles in a CSS file outside of the template for improving browser-side caching.
You could put the file in the static/home/CSS directory

rebeccaTubman: Thank you:) I'll add a file in the directory for my page.

home/tests.py
```diff
@@ -4,97 +4,88 @@
...
-def question_tag_test_data(question_test_data):
-    test_tag = Tag()
-    test_tag.tag_name = 'test_tag_2'
-    test_tag.save()
-    new_pair = Question_Tag()
-    new_pair.question = question_test_data
-    new_pair.tag = test_tag
-    new_pair.save()
-    return test_tag
+class TestManyToManyFeature:
```
ifireball, line 11: this doesn't sound like the best name... names for things in your app should generally be derived from the application model element the functionality is related to

home/tests.py
```diff
@@ -193,3 +184,24 @@ def test_response_context(self, response):
...
+
+class TestTagsPage:
+    @pytest.mark.django_db
+    def test_tags_page_url(self, tags_response):
+        assert tags_response.status_code == 200
+
+    @pytest.mark.django_db
+    def test_tags_page_context(self, tags_response):
+        excepted = '<QuerySet [<Tag: 5th_Grade>, <Tag: Bagrut_Exam>, <Tag: Hebrew>, <Tag: Pitagoras>]>'
+        assert str(tags_response.context['tags']) == excepted
```
ifireball, line 197: its usually a bad idea to use this kind of single-string comparison when you have a more complex data structure. Consider that you might change `__repr__` for various reasons in the future and break the test.

What you should do instead it break down the elements ou want to test - wanna test the type of the object youve got? Use `isinstance`. Wanna test that you got a list of a particular data items? Convert is to a list of simpler objects, like I did in the testing examples I wrote for the message board.

ifireball (approved):
Approving this in the interest of time, because it does seem like you have sufficient test coverage for the feature you've introduced. 

Please see my comments about various elements in the code and apply fixes later.

PR #87, opened 2021-04-29
Add answer feature

### LINKED ISSUE #58: Adding answers to question feature
The feature will allow us to comment on questions.

Tasks:
1. Add text editor to display-question page  (#44 )
2. Create POST request when submitting, and save the comment in Answers table in the db.
3. Change the text editor in admin page.

idokk1:
Enable to add answer with ckeditor.
Included tests for adding new answer .

fix #58


idokk1:
@ifireball do you think I need to move the create profile test to another PR or it's fine here ?

ifireball:
> @ifireball do you think I need to move the create profile test to another PR or it's fine here ?

What do you keep talking about the test, as if you do not also include the code being tested? The important part is the code, not the test... 

In general PRs should include one thing, they can sometimes include more then one thinng if they are small enough.

home/tests.py
```diff
@@ -205,3 +320,33 @@ def tags_response(self, client):
...
         response = client.get(url)
         return response
+
+
+@pytest.mark.django_db
+class TestAddAnswer:
+    def test_add_answer_post(self, client, request, question_test_data, profile):
+        content = {'content': 'test'}
+        request.POST = client.post(f'explore/question_{question_test_data.id}/', data=content)
+        form = CommentForm(request.POST)
```
ifireball, line 330: This is a very strange testing technique, its usually better to test views via the client and not try to call then directly as functions. 
Because you're trying to call the view directly you need to create a whole bunch of stuff to pass to if such as that post object that you're creating by crating the form object... this test is actually leaking a whole bunch of internal implementation details....

Wouldn't it be simpler to have the client perform an HTTP post to the view and then inspect the response? See how your team mates did similar thing in their PRs.

ifireball (changes requested):

idokk1:
@ifireball fixed what you asked.
In addition I added some tests on the html - if form is hidden for visitor and visible for authenticated user ,
if user is not authenticated if the login button is visible for him.
And I expanded the existing tests. 
if you can review please.

rebeccaTubman:
LGTM
Great idea moving the `question_test_data` outside of the class to re-use it.

home/tests.py
```diff
@@ -165,22 +92,16 @@ def question_test_data():
 
 
 @pytest.fixture
-def tag_test_data():
-    tag = Tag(tag_name='test_tag_1')
-    tag.save()
-    return tag
+def profile():
+    profile = Profile.create(username="test_user", password="testtest", email="test@test.com")
+    return profile
```
avivz450, line 97: Great job,
After PR #65 will be merged you can replace these lines with `return request.profile`

home/views.py
```diff
@@ -10,16 +11,21 @@ def landingpage(request):
...
         "answers": question.get_answers_feed(sortAnsBy),
         "answersCount": question.answer_set.count(),
         "tags": question.tags.values(),
-        "title": question.get_question_title()
+        "title": question.get_question_title(),
+        "form": form
     }
+    if request.method == 'POST':
+        profile = Profile.objects.filter(user=request.user).first()
+        return addAnswer(form, profile, question)
```
avivz450, line 28: Great job,
After PR #65  will be merged you can replace these lines with return `addAnswer(form, request.profile, question)`

avivz450 (approved):

danitLevi:
LGTM

danitLevi:
LGTM

idokk1:
@ifireball this PR ready for review
I know this is a big PR, but i don't have an idea how to divide it..
its contains only the answers form with ckeditor, tests for that.

home/tests/test_add_answer.py
```diff
@@ -0,0 +1,76 @@
...
+        # not-authenticated is the div for login button for not authenticated user, should be hidden
+        assert "not-authenticated" not in str(response.content)
+
+    def test_html_not_authenticated_user_form_visible(self, question_test_data, client):
+        response = client.get(f'/explore/question_{question_test_data.id}/')
+        # should be hidden for not authenticated user
+        assert "comment-form" not in str(response.content)
+        # login button for not authenticated user should be visible
+        assert "not-authenticated" in str(response.content)
+
```
ifireball, line 63: why are the test functions below inside another test function?

home/tests/test_question_tag_model_functions.py
```diff
@@ -4,6 +4,25 @@
 from django.utils import timezone
 
 
+@pytest.fixture
```
ifireball, line 7: why is this soperfecial code move needed as part of this PR?

idokk1: I wanted to use this fixture for my tests, so i moved it out of this class instead of copy the entire function.

ifireball:
WRT to the size of the PR - I see 3 separate elements here:
1. The new answer form (I think its about answers but I'm not sure because the variables that refer to them and not always named consistently)
2. The core functionality for creating profiles (I suppose that after this gets in you'll need to refactor alot of old code to use it)
3. The chages to the questions page (some of those look like superficial renaming and moving of thing)

So the mix of those makes it harder to review the Pr and figure out what is relevant. But since its been going for so ling, I suppose  we can keep it like it is. But I did found one thing that looks like a bug in the testing code.

home/tests/test_question_tag_model_functions.py
ifireball: Oh, you don't share fixture like that - to share fixtures you put then in a `conftest.py` file (look it up in the pytest docs).

But in any case you could have written that answer in the PR description and then I wouldn't have had to ask...

idokk1:
create conftest.py and move the creating profile methods to another PR #113

home/tests/test_add_answer.py
idokk1: `test_default_params_answer` 
`test_get_answers`
you are talking about this functions ?

I'm not sure i understand what do you mean.

ifireball: The are indented. So they are inside `test_html_not_authenticated_user_form_visible`. look at your test results, they are probably not running as individual tests.

idokk1: oh, that was by mistake, I will fix that in the next commit when #113  will merge

idokk1:
@ifireball the problems you told me about resolved.

home/tests/test_add_answer.py
```diff
@@ -0,0 +1,73 @@
...
+    assert profile.is_blocked is False
+
+
+@pytest.fixture
+def authenticated_user(client, request):
+    username = 'Lior'
+    password = 'LiorLior'
+    client.login(username=username, password=password)
+    user = authenticate(request, username=username, password=password)
+    return user
```
liornoy: maybe use the authenticated_user fixture that's in conftest.py?

home/tests/test_add_answer.py
```diff
@@ -0,0 +1,73 @@
...
+def profile():
+    profile = Profile.create(username="test_user", password="testtest", email="test@test.com")
+    return profile
+
+
+@pytest.mark.django_db
+def test_default_params_profile(profile):
+    assert profile.user.id >= 0  # check if user saved successfully and get an id
+    assert profile.gender == 'U'  # default gender is U
+    assert profile.is_blocked is False
```
liornoy: what profile.create has to do with testing add answer feature

home/tests/test_add_answer.py
```diff
@@ -0,0 +1,73 @@
+from home.models import Profile, Answer
+import pytest
+from django.contrib.auth import authenticate
+from test_question_tag_model_functions import question_test_data
```
liornoy: question_test_data is a fixture defined in confest.py now not in test_question_tag_model_functions , so maybe this import is not relevant anymore

home/tests/test_question_tag_model_functions.py
```diff
@@ -2,6 +2,25 @@
...
+                        content='Will this question test data pass?',
+                        publish_date=timezone.now(),
+                        subject=subject,
+                        sub_subject=None,
+                        grade='10',
+                        book=None,
+                        book_page=None,
+                        is_edited=False)
+    question.save()
+    return question
```
liornoy: need to remove this as it is now in conftest.py

Pipfile.lock
```diff
@@ -53,6 +53,21 @@
...
+            "index": "pypi",
+            "version": "==6.0.0"
+        },
+        "django-js-asset": {
+            "hashes": [
+                "sha256:8ec12017f26eec524cab436c64ae73033368a372970af4cf42d9354fcb166bdd",
+                "sha256:c163ae80d2e0b22d8fb598047cd0dcef31f81830e127cfecae278ad574167260"
+            ],
+            "version": "==1.2.2"
+        },
```
liornoy, line 70: ckeditor is already merged to the code of master, so maybe rebase

home/tests/test_add_answer.py
```diff
@@ -1,28 +1,5 @@
...
-    assert profile.is_blocked is False
-
-
-@pytest.fixture
-def authenticated_user(client, request):
-    username = 'Lior'
-    password = 'LiorLior'
-    client.login(username=username, password=password)
-    user = authenticate(request, username=username, password=password)
-    return user
```
liornoy: Why didn't you rebased it with the previous commit? I just spent time reviewing a commit that's apparently is not relevant because u made changes in it...

idokk1: I did fixup instead of squash, sorry for that my mistake.. 
fixed it now

rebeccaTubman (commented):
I see you resolved the conflicts. 
Good job! I see test for the model and view functions including redirection. 
I did notice one minor thing - I don't see minimum value for answer content, does that mean you can add a black answer?

idokk1:
> I see you resolved the conflicts.
> Good job! I see test for the model and view functions including redirection.
> I did notice one minor thing - I don't see minimum value for answer content, does that mean you can add a black answer?

for now you do

rebeccaTubman:
> > I see you resolved the conflicts.
> > Good job! I see test for the model and view functions including redirection.
> > I did notice one minor thing - I don't see minimum value for answer content, does that mean you can add a black answer?
> 
> for now you do

OK :)

rebeccaTubman (approved):

danitLevi (approved):

idokk1:
@ifireball the PR is ready for review. Thanks

home/tests/test_add_answer.py
```diff
@@ -0,0 +1,50 @@
...
+        assert "comment-form" in str(response.content)
+        # not-authenticated is the div for login button for not authenticated user, should be hidden
+        assert "not-authenticated" not in str(response.content)
+
+    def test_html_not_authenticated_user_form_visible(self, question_test_data, client):
+        response = client.get(f'/explore/question_{question_test_data.id}/')
+        # should be hidden for not authenticated user
+        assert "comment-form" not in str(response.content)
+        # login button for not authenticated user should be visible
+        assert "not-authenticated" in str(response.content)
```
ifireball, line 36: Missing test for unauthenticated user trying to post

home/views.py
```diff
@@ -12,19 +12,33 @@ def landingpage(request):
     return render(request, 'home/landingpage.html')
 
 
-def displayQuestion(request, **kwargs):
+def displayQuestionPage(request, **kwargs):
```
ifireball, line 15: Tthis is not a Pythonic function name

Also why not define the args with explicit names instead of parsing to opaque `kwargs` structure?

home/views.py
```diff
@@ -12,19 +12,33 @@ def landingpage(request):
...
+    form = CommentForm(request.POST)
     question = Question.objects.all().get(id=kwargs['pk'])
     sortAnsBy = request.GET["sortanswersby"] if 'sortanswersby' in request.GET else ''
```
ifireball, line 18: Can use the dict's `get` method to make this simpler instead of an `if` expression

home/views.py
```diff
@@ -12,19 +12,33 @@ def landingpage(request):
...
         "question": question,
         "answers": question.get_answers_feed(sortAnsBy),
         "answersCount": question.answer_set.count(),
         "tags": question.tags.values(),
-        "title": question.get_question_title()
+        "title": question.get_question_title(),
+        "form": form
     }
+    if request.method == 'POST':
+        return addAnswer(form, request.profile, question)
```
ifireball, line 28: not sure moving the code out to a function makes sense, and the risk is that people migt think that helper function is a view function since its in the views file.

The function name is also not Pythonic.

ifireball (changes requested):

home/views.py
idokk1: I added some functions to this function, because my feature is in the same view as Lior's feature.
the use of kwargs is was on his PR

home/views.py
idokk1: same here, this is content of Lior's PR.

home/views.py
ifireball: 1. Blaming your friend is not nice.
2. If its broken - fix it.

home/views.py
idokk1: fixed

home/tests/test_add_answer.py
idokk1: added

home/views.py
idokk1: not blaming him, just afraid to touch his code (that already merged) 
but I'll do it

ifireball: But you did touch it... Also why fear, you have tests no? Don't you trust the tests to tell you if you've broken anything important?

home/views.py
idokk1: fixed

home/views.py
idokk1: you right, fixed.

idokk1:
@ifireball I fixed the last reviews and resolved the conflicts.

home/tests/test_add_answer.py
```diff
@@ -0,0 +1,55 @@
...
+    def test_get_answers(self):
+        out = Answer.get_answers_by_date()
+        assert all(isinstance(a, Answer) for a in out)
+        assert set([
+            (2, 2, 'IDK'),
+            (1, 1, 'pretty sure its 2 but I suggesting you to check with another resources '),
+        ]).issubset(list(out.values_list('profile', 'question', 'content')))
+
+    def test_post_answer_not_authenticated_user(self, client, question_test_data, comment_content):
+        with pytest.raises(IntegrityError):
```
ifireball, line 54: I'm pretty sure this is supposed to return some kind of an HTTP error code and possibly display some error page, not raise an exception.

home/views.py
```diff
@@ -12,16 +12,29 @@ def landingpage(request):
     return render(request, 'home/landingpage.html')
 
 
-def displayQuestion(request, **kwargs):
-    question = get_object_or_404(Question, id=kwargs['pk'])
-    sortAnsBy = request.GET["sortanswersby"] if 'sortanswersby' in request.GET else ''
+def display_question_page(request, pk):
+    def add_answer(form, profile, question):
```
ifireball, line 16: what is the point of having a nested function like this, it seems to be just adding some unnecessary complexity to the code...

idokk1: removed

home/tests/test_add_answer.py
idokk1: but unauthenticated cannot see the form to add answer at all (and I checked it).
and if someone try to post to this page somehow without being authenticated user it will fail.
it's not enough?

ifireball: "seeing" the form is not required for doing an HTTP POST request - remember that the browser and the server are connected via an untrusted network.

You need to check that your site is generating the failure as needed, otherwise you have a security issue.

This test is using the client to do an HTTP request to the site. When you are doing an HTTP request to a site you are supposed to get some kind of an HTTP response (Errors are valid responses too). If you get an exception from the client it means it somehow failed to communicate.

Also, are you sure you need all the fixtures for this test?

idokk1: added a check that if there is no user , raise 404. and created test for that.

rebeccaTubman (approved):

home/views.py
```diff
@@ -13,16 +14,29 @@ def landingpage(request):
...
         "answersCount": question.answer_set.count(),
         "tags": question.tags.values(),
-        "title": question.get_question_title()
+        "title": question.get_question_title(),
+        "form": form
     }
+    if request.method == 'POST':
+        if form.is_valid():
+            if (request.profile is None):
+                raise Http404()
```
liornoy, line 32: Maybe switch it to 401 response - unauthorized. (Barak recommended it to me also in a similar case)
you can see how I did it in my PR in views.py

home/tests/test_add_answer.py
```diff
@@ -0,0 +1,54 @@
...
+        out = Answer.get_answers_by_date()
+        assert all(isinstance(a, Answer) for a in out)
+        assert set([
+            (2, 2, 'IDK'),
+            (1, 1, 'pretty sure its 2 but I suggesting you to check with another resources '),
+        ]).issubset(list(out.values_list('profile', 'question', 'content')))
+
+    def test_post_answer_not_authenticated_user(self, client, question_test_data, comment_content):
+        response = client.post(f'/explore/question_{question_test_data.id}/', data=comment_content)
+        assert response.status_code == 404
```
liornoy, line 54: Maybe switch response to unauthorisde, see my comment below

home/views.py
idokk1: you right. 404 is really doesn't the right status code for this case

home/templates/home/question_detail.html
```diff
@@ -44,7 +44,7 @@ <h3> {{ question.title }} </h3>
 	<div class="answer-display mt-1 col-md-10">
 	<div class="row">
 		<div class="asnwer-conent col-md-10">
-			<p>{{ answer.content }}</p>
+			<p>{{ answer.content | safe }}</p>
```
ifireball, line 47: This may or may not be a security issue - does the RichTextField class perfom any HTML sanitation to ensure there are not things like `<script>` tags?

ifireball (changes requested):

home/templates/home/question_detail.html
idokk1: @ifireball fixed it in #131 , wait for merge and then i will use the same solution here.

idokk1: fixed

idokk1:
@ifireball , fixed the security issue and resolved the rest of the reviews comments.

home/forms.py
```diff
@@ -19,3 +20,11 @@ class SignUpForm(UserCreationForm):
     class Meta:
         model = User
         fields = ('username', 'first_name', 'last_name', 'email', 'gender', 'password1', 'password2', )
+
+
+class CommentForm(forms.ModelForm):
+    content = RichTextField(blank=True, null=True)
```
ifireball, line 26: Should this be the `RichTextBleackFeild` too? Also id expect to see `| safe` somewhere as a result of this change.

idokk1: I changed it, 
the important thing is that `RichTextBleackField ` will be on the model , because there I preform the validation. 
the  `RichTextBleackField ` is same as `RichTextField` but with the validations.
and there is `|safe` on the question_detail.html ( `<p>{{ answer.content | safe }}</p>` )

idokk1:
the PR is ready for re-review @ifireball

ifireball (approved):

PR #91, opened 2021-05-03
Adding idea folder to git ignore file

danitLevi:
I usually use PyCharm .PyCharm automatically creates ".idea" folder. 
I add this PR for my own use so I won't add this folder into this repository by mistake .

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -133,3 +133,6 @@ dmypy.json
 
 #nohup.out
 nohup.out
+
+# idea
+.idea/


Idokah:
I have this weird folder to.
what is your use of it?

danitLevi:
> I have this weird folder to.
> what is your use of it?

@Idokah 

I usually use PyCharm .PyCharm automatically creates ".idea" folder that contains IntelliJ’s project specific settings files. 
Maybe you use PyCharm too ? 
In addition, you can read more about this folder it in the following link :
https://rider-support.jetbrains.com/hc/en-us/articles/207097529-What-is-the-idea-folder-

ifireball:
Why not commit the folder to Git to share the PyCharm configuration between you?

danitLevi:
> Why not commit the folder to Git to share the PyCharm configuration between you?

@ifireball 
I use PyCharm only as a text editor( like notepad++) because it is more convenient for me. Then, I run the code via a virtual machine with Vagrant. So I thought there is no need for the "idea" folder in GitHub. Does it make sense?

ifireball:
> > Why not commit the folder to Git to share the PyCharm configuration between you?
> 
> @ifireball
> I use PyCharm only as a text editor( like notepad++) because it is more convenient for me. Then, I run the code via a virtual machine with Vagrant. So I thought there is no need for the "idea" folder in GitHub. Does it make sense?

Up to you, it can take some work to get PyCharm to work properly with Pipenv, and it can also be a quite ueseful Git front-end. The debugger could be made to connect to the VM, but the commercial version is required for that AFAIK. In any case its a team decision wither you want to share this configuration or not. IIRC PyCharm does suggest to add the files to git by default.

avivz450:
I think we don't need this folder in GitHub and we can just add it to the .gitignore as done in this PR

avivz450 (approved):

idokk1 (approved):

ifireball (approved):

PR #93, opened 2021-05-03
Add timezone specification to all datetime

### LINKED ISSUE #92: Add timezone support to all test-data
As for now, we use "naive" datetime for our test data.
for example: datetime(2019, 5, 3, 9, 40, 35)

This causes Django to send warnings when using makemigrations and when using pytest.

Fix for it is adding "tzinfo=pytz.UTC".

liornoy:
This will remove the warnings Django throws when running migrate and tests

modified:   home/migrations/0013_add_questions_test_data.py
modified:   home/migrations/0015_add_answers_test_data.py
modified:   home/tests.py

fix #92

### DIFF
--- a/home/migrations/0013_add_questions_test_data.py
+++ b/home/migrations/0013_add_questions_test_data.py
@@ -1,5 +1,6 @@
 from django.db import migrations, transaction
 from datetime import datetime
+import pytz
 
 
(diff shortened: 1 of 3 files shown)


danitLevi:
LGTM

ifireball (approved):

PR #98, opened 2021-05-03
Adding logout page

### LINKED ISSUE #97: Add logout option
add logout option by Django

idokk1:
Add Django built in logout option.
Similar to login ,because it's implemented by Django I didn't add tests for that  except for test approaching /logout url.
fix #97

### DIFF
--- a/home/tests/test_logout.py
+++ b/home/tests/test_logout.py
@@ -1,0 +1,28 @@
+import pytest
+
+
+@pytest.mark.django_db
+class TestLogout:
+    def test_status_code(self, client):
+        response = client.get('/logout/')
+        assert response.status_code == 302  # redirected
+
+    def test_redirection_default_url(self, client):
+        response = client.get('/logout/')
+        assert response.url == '/'  # redirected to home page.
+
+    def test_logout(self, client, authenticated_user):
+        response = client.get('/')
+        assert response.wsgi_request.user.is_authenticated  # successfully logged in
+        response = client.get('/logout/')
+        assert not response.wsgi_request.user.is_authenticated   # successfully logged out
+
+    def test_unauthenticated_user_view(self, client):
+        response = client.get('/')
+        assert "Login" in str(response.content)
+        assert "Logout" not in str(response.content)
+
+    @pytest.mark.parametrize(('next'), [('about'), ('explore'), ('tags')])
+    def test_redirection_to_last_page_url(self, client, request, next):
+        response = client.get('/logout/?next=/{0}/'.format(next))
+        assert response.url == '/{0}/'.format(next)
--- a/home/templates/home/base.html
+++ b/home/templates/home/base.html
@@ -54,7 +54,7 @@
 			  <ul class="nav navbar-nav" >
 				<li><p class="nav-item m-2"><b>Hello {{user.username}}</b></p></li>
 				<li><a class="nav-item nav-link m-2 " href="#"><b>Profile</b></a></li>
-				<li><a class="nav-item nav-link m-2" href="#"><b>Logout</b></a></li>
+				<li><a class="nav-item nav-link m-2" href="{% url 'logout' %}?next={% firstof request.path '/' %}"><b>Logout</b></a></li>
 			  </ul>
 			  {% else %}
 			  <p class="nav-item nav-link m-2"><b>Hello visitor</b></p>
--- a/home/urls.py
+++ b/home/urls.py
@@ -1,7 +1,7 @@
 from django.urls import path
+from django.contrib.auth import views as auth_views
 from . import views
 from .views import QuestionsListView
-from django.contrib.auth import views as auth_views
 
 urlpatterns = [
     path('about/', views.about, name='about'),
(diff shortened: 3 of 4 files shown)


ifireball (changes requested):
Needs more tests like the other PR.

idokk1:
> Needs more tests like the other PR.

Added

avivz450:
LGTM

home/templates/home/logout.html
```diff
@@ -0,0 +1,14 @@
+{% extends "home/base.html" %}
+{% block content %}
+<body class="main-bg">
+    <h1>Logged Out</h1>
+
+    <script>
```
ifireball, line 6: IIRC you can do this with an HTTP header, so you don't have to trust JS on the client side.

https://en.wikipedia.org/wiki/URL_redirection#Refresh_Meta_tag_and_HTTP_refresh_header

But it is considered a UI anti-pattern to have a flashing disappearing page like this, its better to just redirect to the main page. Users usually can tell they are logged out because their name or profile icon no longer appears on the navigation bar.

home/templates/home/logout.html
```diff
@@ -0,0 +1,10 @@
+{% extends "home/base.html" %}
+{% block content %}
+<body class="main-bg">
+    <h1>You are logged out.</h1>
+    <p>
+        See you soon :( <br>
+        Click <a href="/">here</a> to return the home page.
```
idokk1, line 7: @ifireball this is fine?

home/templates/home/logout.html
ifireball: It might be good enough to redirect immediately. It seems the LogoutView has the built-in ability to do this using the `next_page` attribute or the [`LOGOUT_REDIRECT_URL`](https://docs.djangoproject.com/en/3.2/ref/settings/#std:setting-LOGOUT_REDIRECT_URL) setting.

https://docs.djangoproject.com/en/3.2/topics/auth/default/#all-authentication-views

You can also try to be a bit more clever and have the Login and Logout views redirect the users back to where they came from if possible, the can provide a better experience when the login/logout button are on a navigation bar that is seen on every page.

home/tests.py
```diff
@@ -476,3 +476,32 @@ def test_authenticated_user_view(self, client, valid_user_details):
...
+@pytest.mark.django_db
+class TestLogout:
+    def test_status_code(self, client):
+        response = client.get('/logout/')
+        assert response.status_code == 302
+
+    def test_redirection_url(self, client, request):
+        # each page transfer his own url as a next parameter
+        response = client.get('/logout/?next=/about/')
+        assert response.url == '/about/'  # redirected to home page.
```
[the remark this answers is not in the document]
rebeccaTubman: Ok now I understand. Thanks:)

idokk1:
> Lior's PR that reorganized the tests was merged. You can create a new test file in the folder for your tests to resolve the conflicts.

Done, Thank you

idokk1:
@ifireball this PR ready for review

home/tests/test_logout.py
```diff
@@ -0,0 +1,29 @@
...
+
+
+@pytest.mark.django_db
+class TestLogout:
+    def test_status_code(self, client):
+        response = client.get('/logout/')
+        assert response.status_code == 302  # redirected
+
+    def test_redirection_default_url(self, client):
+        response = client.get('/logout/')
```
ifireball, line 11: could be merged with the test above

home/tests/test_logout.py
```diff
@@ -0,0 +1,29 @@
...
+
+    def test_unauthenticated_user_view(self, client):
+        response = client.get('/')
+        assert "Login" in str(response.content)
+        assert "Logout" not in str(response.content)
+
+    def test_redirection_url(self, client, request):
+        # each page transfer his own url as a next parameter
+        response = client.get('/logout/?next=/about/')
+        assert response.url == '/about/'  # redirected to home page.
```
ifireball, line 29: the comment is lying, why do you need it anyway?

home/tests/test_logout.py
```diff
@@ -0,0 +1,29 @@
...
+        response = client.get('/logout/')
+        assert not response.wsgi_request.user.is_authenticated   # successfully logged out
+
+    def test_unauthenticated_user_view(self, client):
+        response = client.get('/')
+        assert "Login" in str(response.content)
+        assert "Logout" not in str(response.content)
+
+    def test_redirection_url(self, client, request):
+        # each page transfer his own url as a next parameter
```
ifireball, line 27: could you explain this via the test name instead of a comment?

ifireball (changes requested):
found some small issues with the code structure

home/tests/test_logout.py
idokk1: oops

idokk1:
fixed

ifireball (approved):

ifireball:
not seeing any reviews and approvals by team mates

home/tests/test_logout.py
```diff
@@ -0,0 +1,28 @@
...
+        assert response.wsgi_request.user.is_authenticated  # successfully logged in
+        response = client.get('/logout/')
+        assert not response.wsgi_request.user.is_authenticated   # successfully logged out
+
+    def test_unauthenticated_user_view(self, client):
+        response = client.get('/')
+        assert "Login" in str(response.content)
+        assert "Logout" not in str(response.content)
+
+    def test_redirection_to_last_page_url(self, client, request):
```
rebeccaTubman, line 26: I'm not sure it's needed, but you could test the redirection after logout works on a few pages using parametrize.
```
response = client.get('/logout/?next=/{0}/'.format(check))
assert response.url == '/{0}/'.format(check)
```

idokk1: good idea, added

rebeccaTubman (approved):

danitLevi:
LGTM

danitLevi (approved):

PR #112, opened 2021-05-11
Added redirections to 'Insert a question' page + tests

### LINKED ISSUE #108: Add redirections at question insertion page
2 redirections :
1. If the user is not logged in, redirection to the login page.
2. After submitting a question, redirection to that question page.

avivz450:
- Entered 2:
1. When the user is not logged in once he clicked on "Insert a question".
In that case, he will be navigated to the login page.
2. After inserting a question, redirection to the question page.
The redirection will be only if the input is valid.

- Added tests for these two

fix #108

### DIFF
--- a/home/tests/test_insert_question_feature.py
+++ b/home/tests/test_insert_question_feature.py
@@ -99,20 +99,22 @@ class TestInsertQuestionFeature:
             # User entered all the required fields and sub-subject
             ({'title': "Question in Math",
               'subject': 1, 'grade': Grade.GRADE7, 'book': 2}),
-            # User entered all the required fields and sub-subject and book
+            # User entered all the required fields, sub-subject and book
             ({'title': "Question in Math",
               'subject': 1, 'grade': Grade.GRADE7, 'book_page': 23}),
             # User entered all the required fields and a valid book page
         ])
         @pytest.mark.django_db
-        def test_post_valid_question_with_client(self, client, valid_data):
+        def test_post_valid_question_with_client(self, client, valid_data, authenticated_user):
             with pytest.raises(Question.DoesNotExist):
                 assert Question.objects.get(title=valid_data["title"])
 
-            client.login(username='Lior', password='LiorLior')
-            client.post('/explore/new_question', data=valid_data)
+            response = client.post('/explore/new_question', data=valid_data)
 
             assert Question.objects.filter(title=valid_data["title"]).exists()
+            assert response.status_code == 302
+            question_id = Question.objects.get(title=valid_data["title"]).id
+            assert response.url == "/explore/question_" + str(question_id) + "/"
 
         @pytest.mark.parametrize("invalid_data", [
             ({'title': "Question in Math", 'content': 'How much is it 1+1?', 'grade': Grade.GRADE7}),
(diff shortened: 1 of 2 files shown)


home/tests/test_insert_question_feature.py
```diff
@@ -110,9 +110,12 @@ def test_post_valid_question_with_client(self, client, valid_data):
                 assert Question.objects.get(title=valid_data["title"])
 
             client.login(username='Lior', password='LiorLior')
-            client.post('/explore/new_question', data=valid_data)
+            response = client.post('/explore/new_question', data=valid_data)
```
rebeccaTubman, line 114: All the tests below seem important.
The question is whether it is possible to split them into several functions because we have received comments in the past about the amount of asserts in that function.

avivz450: All of these asserts are related to the process of posting a valid question:
Once the user has entered a question with valid input, the post should be saved in the database and the user should be automatically navigated to the question page he posted. I can split this test into 2 tests that check if the question exists in the db and if a redirection will be made, but it seems to me that in this case, it is a duplication of code.
Besides, I have previously received a comment from Barak on merging the tests in these cases (in this [PR](https://github.com/beyond-io/shlifim/pull/65)).

liornoy:
Maybe add a test to where the page is redirected if an invalid question was submited?

avivz450:
> Maybe add a test to where the page is redirected if an invalid question was submited?

He's not supposed to be redirected. 
After this PR will be merged, I will do another one that will show an 'alert' message once the details are incorrect

home/tests/test_insert_question_feature.py
```diff
@@ -110,9 +110,12 @@ def test_post_valid_question_with_client(self, client, valid_data):
                 assert Question.objects.get(title=valid_data["title"])
 
             client.login(username='Lior', password='LiorLior')
```
idokk1, line 112: you can use the fixture authenticated_user instead of preform client.login in this and in the next test function.
you dont need to import anything because its in the conftest.py file, just add a parameter to your function : 
authenticated_user and the client will be already logged in. like in Answer feature tests right [here](https://github.com/beyond-io/shlifim/pull/87/files#diff-f7c34089a18939177812719101f6af75ca1bd7ed354d3fc4bf8ea6509985bc89R17)

liornoy (approved):

home/tests/test_insert_question_feature.py
avivz450: Thanks for the offer. 
I can do it, but I didn't understand why should I prefer that suggestion instead of what I did?

rebeccaTubman: I agree with Ido, if you have a fixture that does the job for you why do it manually again? 
The whole point of conftest.py was to enable us to use in fixtures in different test files.

avivz450: Ok, added

rebeccaTubman (approved):
Good job!

home/tests/test_insert_question_feature.py
```diff
@@ -99,20 +99,22 @@ class TestViews:
...
             with pytest.raises(Question.DoesNotExist):
                 assert Question.objects.get(title=valid_data["title"])
 
-            client.login(username='Lior', password='LiorLior')
-            client.post('/explore/new_question', data=valid_data)
+            response = client.post('/explore/new_question', data=valid_data)
 
             assert Question.objects.filter(title=valid_data["title"]).exists()
+            assert response.status_code == 302
+            question_id = Question.objects.filter(title=valid_data["title"])[0].id
```
danitLevi, line 116: @avivz450 
I think you can use "get" instead of "filter" and [0]. 
I mean you can write 
` question_id = Question.objects.get(title=valid_data["title"]).id`
instead of 
` question_id = Question.objects.filter(title=valid_data["title"])[0].id`

avivz450: Good idea, done

home/tests/test_insert_question_feature.py
```diff
@@ -125,26 +127,30 @@ def test_post_valid_question_with_client(self, client, valid_data):
             # User entered all the required fields and to high book page
         ])
         @pytest.mark.django_db
-        def test_post_invalid_question_with_client(self, client, invalid_data):
-            client.login(username='Lior', password='LiorLior')
+        def test_post_invalid_question_with_client(self, client, invalid_data, authenticated_user):
             client.post('/explore/new_question', data=invalid_data)
```
ifireball, line 131: Its not directly related to this PR - but it seems that assertions for the HTTP response are missing here and in other similar tests. How do you know the right error message is shown to the user?

ifireball (approved):

PR #113, opened 2021-05-13
Creating conftest.py

idokk1:
Creating conftest.py and put there useful common pytest fixtures . 
Add create method for Profile model.
move question_test_data() fixture from test_question_tag_model_function.py to conftest.py.
fix #114

PR #115, opened 2021-05-13
Fix like and dislike buttons in display question page

### LINKED ISSUE #89: Make thumb up/down in display question page work

liornoy:
All the changes are detailed in the commits comments,
Also, see the issue below to see how the page looks with the buttons
fix #89


home/tests/test_display_question_feature.py
```diff
@@ -107,10 +91,97 @@ def test_response_context(self, response):
...
+        @pytest.mark.django_db
+        def test_login_link_shown_for_unsigned_user(self, client):
+            """
+            Tests if the page loads the link "login to like" for unsigned users
+            """
+            url = reverse('question-detail', args=[14])
+            display_question_response = client.post(url)
+            char_content = display_question_response.content.decode(display_question_response.charset)
+            assert 'Login</a> to like</small>' in char_content
+            assert 'disabled' in char_content
```
rebeccaTubman, line 187: Great test coverage - like\unlike and dislike\undislike logged in users, disable content fo unsigned user and invalid inputs.

home/urls.py
```diff
@@ -10,5 +10,6 @@
     path('tags/', views.tags, name='tags'),
     path('explore/', QuestionsListView.as_view(), name='explore-page'),
     path('explore/new_question', views.new_question, name='new_question'),
-    path('login/', auth_views.LoginView.as_view(template_name='home/login.html'), name='login')
+    path('login/', auth_views.LoginView.as_view(template_name='home/login.html'), name='login'),
+    path('thumbs', views.thumbs, name='thumbs')
```
rebeccaTubman, line 14: Just to make sure I understand, you made a different path for thumbs so you could call the function in views for each answer, right?

home/tests/test_display_question_feature.py
rebeccaTubman, line 187: Great test coverage - like\unlike and dislike\undislike logged in users, disable content fo unsigned user and invalid inputs.

home/views.py
```diff
@@ -14,17 +16,35 @@ def landingpage(request):
 
 def displayQuestion(request, **kwargs):
     question = Question.objects.all().get(id=kwargs['pk'])
-    sortAnsBy = request.GET["sortanswersby"] if 'sortanswersby' in request.GET else ''
+    sort_answer_by = request.GET["sortanswersby"] if 'sortanswersby' in request.GET else ''
```
danitLevi, line 19: I think instead of doing if statement  you can just write here the following line
` sort_answer_by = request.GET["sortanswersby", '']`

home/tests/test_display_question_feature.py
danitLevi, line 187: Great tests .

home/tests/test_display_question_feature.py
```diff
@@ -107,10 +91,97 @@ def test_response_context(self, response):
...
-        def test_sorting_answers(self, client, filterType, expected):
-            url = f'/explore/question_14/?sortanswersby={filterType}'
+        def test_sorting_answers(self, client, filter_type, expected):
+            url = f'/explore/question_14/?sortanswersby={filter_type}'
             response = client.get(url)
-            answer = response.context['answers'].first()
+            answer = response.context['answers_structure'][0][0]
             assert str(answer.content) == expected
+
+    class TestLikesMechanism:
```
danitLevi, line 102: Great tests !

home/tests/test_display_question_feature.py
```diff
@@ -107,10 +91,97 @@ def test_response_context(self, response):
...
+            assert answerA in str(response.context['answers_structure'])
+
+        @pytest.mark.django_db
+        def test_invalid_thumbs_call_from_unsigned_user(self, client):
+            """
+            Unsigned user should not be able to like an answer
+            """
+            url = reverse('thumbs')
+            response = client.post(url, {'button': "like 9"})
+            assert response.status_code == 404
```
avivz450, line 164: Maybe you should add redirection to a log-in page in this situation instead of to the error page.
What do you think?

home/tests/test_display_question_feature.py
```diff
@@ -107,10 +91,97 @@ def test_response_context(self, response):
...
+            in this test, a logged user press dislike on (answer A) which he disliked before
+            and make it un-disliked
+            """
+            user = Profile.objects.all().get(user_id=1).user
+            client.force_login(user)
+            url = reverse('thumbs')
+            response = client.post(url, {'button': "dislike 9"})
+            response = client.get(response.url)
+            answerA = '(<Answer: Answer A>, \'bi bi-hand-thumbs-up\', \'bi bi-hand-thumbs-down'
+            assert answerA in str(response.context['answers_structure'])
```
avivz450, line 155: How do you know what were the choices of the user before in these 4 methods?

idokk1:
I see that in a lot of tests you are fetching user from DB and then login. 
In my last PR (#113 ) I added the file conftest.py that there are some useful fixtures there (you are invited to add some if you think that they will be relevant to the rest of us) , one of them is 
`authenticated_user ` , that is do just that (fetching and login user). you don't need to import anything just add to your function parameters the parameter: authenticated_user. 
then the client will be already logged in and you can test the rest of the logic.
you can se examples to the use of this fixture [here](https://github.com/beyond-io/shlifim/pull/87/files#diff-f7c34089a18939177812719101f6af75ca1bd7ed354d3fc4bf8ea6509985bc89R17)

home/urls.py
liornoy: yes

home/views.py
liornoy: Thanks I fixed it

home/tests/test_display_question_feature.py
liornoy: I think I don't want to do it because:
The thumbs view should be reached only from a legit like/dislike button press of a signed user.
By redirecting to the login page when entering a URL that shouldn't be entered manually is a way of saying - what you did is a legit and acceptable operation, and let me help you sign in so you could complete this operation. but I don't want to allow it. Moreover, this test is testing a case that will never happen because the thumbs view needs to get an answer ID from a post request, so this is another reason it should fail.
Unsigned users have a link for login under the disabled like buttons, as you can see in the picture in the issue linked to this PR.

home/tests/test_display_question_feature.py
liornoy: By the "answer_test_data" variable of 0015_add_answers_test_data.py file

liornoy:
> I see that in a lot of tests you are fetching user from DB and then login.
> In my last PR (#113 ) I added the file conftest.py that there are some useful fixtures there (you are invited to add some if you think that they will be relevant to the rest of us) , one of them is
> `authenticated_user ` , that is do just that (fetching and login user). you don't need to import anything just add to your function parameters the parameter: authenticated_user.
> then the client will be already logged in and you can test the rest of the logic.
> you can see examples to the use of this fixture [here](https://github.com/beyond-io/shlifim/pull/87/files#diff-f7c34089a18939177812719101f6af75ca1bd7ed354d3fc4bf8ea6509985bc89R17)

Thanks for the comment and suggestion.
In my case, I need to use Rebecca's user for the specific likes/dislikes test data,
Also, I'm using the client.force_login instead of client.login, as suggested by Barak, and from reading about it, this way the tests are less time-consuming because force_login does login quicker by skipping all the authentication part.

avivz450 (approved):

rebeccaTubman (approved):

danitLevi (approved):

home/models.py
```diff
@@ -100,6 +101,24 @@ class Meta:
     def __str__(self):
         return f"Question #{self.id} : {self.title} ( {self.publish_date.strftime('%d/%m/%Y %H:%M')} )"
 
+    def create_answers_structure(self, filter_type, current_profile):
```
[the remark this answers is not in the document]
ifireball: Having, looked at this again - I now understand this is a querying function that returns a list of answers. The function name is a bit misleading beucause I usually interpret "structure" to mean something like "struct" in C. 

Also this being an object-level method, one expects it to return object-level data and not a full list of objects.

It seems like the purpose of this is to get icons for display in the UI. The same could be achieved by adding a couple of property getters to the answers object instead of a whole wrapper to the `get_answers_feed` class method.

But this is mixing UI concerns such as icon name with model code - I would advise to keep the UI-specific stuff in the UI code. The model can still have helper functions to make it easy to know if an answer is liked or not, but things like CSS class names belong in the UI layer.

home/models.py
```diff
@@ -139,23 +158,35 @@ class Answer(models.Model):
...
-    def thumb_up_answer(self):
-        self.likes_count = self.likes_count + 1
-
-    def thumb_down_answer(self):
-        self.dislikes_count = self.dislikes_count + 1
-
     def set_is_edited(self, newVal):
         self.is_edited = newVal
 
+    def update_thumbs(self, profile, thumb_type):
```
ifireball, line 172: I would make thumb_type an enum of just split this into two separate methods.

home/tests/test_display_question_feature.py
```diff
@@ -97,8 +79,10 @@ def test_response_context(self, response):
             '''
             expectedPairs = [
                 ('question', 'Question #14 : g forwards, it was even later than ( 28/04/2021 23:14 )'),
-                ('answers', '<QuerySet [<Answer: Answer B>, <Answer: Answer A>,\
- <Answer: Popular Answer>, <Answer: Old Answer>]>'),
```
ifireball, line 101: please maintain consistent code indentation

home/tests/test_display_question_feature.py
```diff
@@ -107,10 +91,97 @@ def test_response_context(self, response):
...
+    class TestLikesMechanism:
+        @pytest.mark.django_db
+        def test_logged_user_like(self, client):
+            """
+            in this test, a logged user press like on (answer A) which he disliked before
+            """
+            user = Profile.objects.all().get(user_id=1).user
+            client.force_login(user)
+            url = reverse('thumbs')
+            response = client.post(url, {'button': "like 9"})
```
ifireball, line 111: Missing assertion about the response from the post reqeuest, you seem to be blindly assuming it will redirect you

home/tests/test_display_question_feature.py
```diff
@@ -107,10 +91,97 @@ def test_response_context(self, response):
...
+
+        @pytest.mark.django_db
+        def test_logged_user_dislike(self, client):
+            """
+            in this test, a logged user press dislike on (answer B) which he liked before
+            """
+            user = Profile.objects.all().get(user_id=1).user
+            client.force_login(user)
+            url = reverse('thumbs')
+            response = client.post(url, {'button': "dislike 10"})
```
ifireball, line 124: Missing assertion about the response from the post reqeuest, you seem to be blindly assuming it will redirect you

home/tests/test_display_question_feature.py
ifireball: It would probably be better to respond with something like 401 (unauthorized) or 403 (forbidden) then 404 (not found):

https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/401

home/tests/test_display_question_feature.py
ifireball: Nope, missing garbage intput use cases. If you had any you would have found the security vulnerabilities you might currently have in the view code.

home/views.py
```diff
@@ -14,17 +16,35 @@ def landingpage(request):
...
+        "title": question.get_question_title(),
+        "is_user_logged_in": request.user.is_authenticated
     }
     return render(request, 'home/question_detail.html', context)
 
 
+def thumbs(request):
+    if request.user.is_authenticated:
+        button = str(request.POST.get('button'))
+        (thumb_type, answer_id) = button.split(' ')
```
ifireball, line 34: this is making too many assumption about the button string in the form post data - who said its even splittable? not to mention containing the right "thumb" string? it could be total garbage...

home/templates/home/question_detail.html
```diff
@@ -71,28 +66,23 @@ <h3> {{ question.title }} </h3>
...
+			</button>
+			  - {{ answer.likes.count }}
+			  <button type="submit", name="button", value="dislike {{ answer.id }}", class="btn btn-danger btn-sm" 
+			  {% if not is_user_logged_in %}disabled {% endif %}>
+			 <i class="{{ disliked_icon }}"></i>
+			  </button>
+			  - {{ answer.dislikes.count }}
+			  {% if not is_user_logged_in %}
+				<small><a href="{% url 'login' %}">Login</a> to like</small>
+			 {% endif %}
```
ifireball, line 85: missing form closure tag?

home/templates/home/question_detail.html
```diff
@@ -71,28 +66,23 @@ <h3> {{ question.title }} </h3>
...
-            </svg>
-          </a>
+		  <form action="{% url 'thumbs'%}" method="POST">
+			 {% csrf_token %}
+			<button type="submit", name="button", value="like {{ answer.id }}", class="btn btn-primary btn-sm"
```
ifireball, line 73: could include the answer ID in the action URL or as a hidden field.

home/views.py
```diff
@@ -14,17 +16,35 @@ def landingpage(request):
...
+        answer = get_object_or_404(Answer, id=answer_id)
+        profile = request.profile
+        answer.update_thumbs(profile, thumb_type)
+        question_id = answer.question.id
+
+        # HTTP_REFERER will return the url the user came from.
+        # if previous url doesn't exist (tests / incognito) then will return the second argument url
+        responseURL = request.META.get('HTTP_REFERER', reverse('question-detail', args=[question_id]))
```
ifireball, line 42: You could include the qusetion ID and answer ID in the URL to avoid having to trust `HTTP_REFERER` which can be blocked for various reasons (Because evil companies use this for spying). 

You can even make the thumb status be a part of the HTTP request and have the the Django routing mechanism do all the data validation and parsing  for you - the nice thing about Django's `url` funciton is that it can get additional arguments to build more complex URLs.

(Dong the whole thing in a GET request might expose it to CSRF attacks, not sure if that is a huge risk, but it can be mitigated by forcing it to be POST and include the CSRF token - the action URL of a form can be as complex as you want..)

ifireball (changes requested):

home/tests/test_display_question_feature.py
liornoy: I don't understand, what kind of garbage input do you mean? is it the value of the 'button' argument in the post request? do you want me to check if someone creates a mischief request with bad value in it? even though it's an action a normal user can't do even by mistake?

home/views.py
liornoy: The reason I used HTTP_REFERER is not that I wanted to go back to the page displaying this question, but to get the exact URL that includes the query param the user had, for example "?sortanswersby=votes&page#answers" filtering by votes and scrolling down to answers part. I can't find a solution how to solve it otherwise

ifireball: You can reproduce the query parameters because you know all of them when you generae the page. WRT the anchor navigation (The `#` part), I'm not sure how important is that really if the answers list can be lnger then a single screen. It may be more useful to unconditionally focus the user's browser on the particular answer they "liked".

You can use `HTTP_REFERER` for extra optimization, but you must validate it like the untrustworthy piece of input that it is. Blindly redirecting users to the URL in `HTTP_REFERER` means your side can be used and an HTTP attack proxy.

liornoy: I think I'll leave this optimization alone for now

ifireball: This is a security/input validation issue not an optimization.

home/tests/test_display_question_feature.py
liornoy: fixed

home/templates/home/question_detail.html
liornoy: fixed

home/tests/test_display_question_feature.py
liornoy: fixed, not using form and post anymore

home/models.py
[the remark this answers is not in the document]
liornoy: I changed it as you suggested. Now having a tuple array of:
(answer object, bool is_user_liked, bool is_user_disliked)
I use the bools to pick which icon to display in the template

home/models.py
```diff
@@ -115,16 +134,16 @@ def add_tags_to_question(self, tags):
                 new_pair.tag = tag
                 new_pair.save()
 
-    def get_answers_feed(self, filterType=''):
+    def get_answers_feed(self, filter_type=''):
```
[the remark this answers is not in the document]
liornoy: ok thanks, will watch for that from now on

home/models.py
liornoy: I decided to split into 2 methods

home/tests/test_display_question_feature.py
liornoy: fixed

home/views.py
liornoy: fixed - not using form post any more

home/views.py
liornoy: I mean I'll leave the use of HTTP_REFERER  for now, will handle it later - opened issue #128  for it.

ifireball: Thanks for creating a teachable moment. This is precisely how not to defer wort to the future. Look at #128, what do you think are the chances of it ever getting prioritized? Close to zero right? Because project managers and user representative almost never care about invisible elements like this.

You can never defer security and tests work to the future, when you do:
1. Your users are at risc of relying on features that might break due to insufficient testing
2. Your users are at risc of security attacks.

So the only kind of work you can defer to the future are *user-visible features*. In this case, the right trade-off is to make the redirection a little less smart (For example, redirect to the top of the page), and defer the usage of the header to the future.

liornoy: First, thank you for the lesson. I understand now the importance of dealing with security/testing issues urgently.
I'm currently returning respond to - reverse('question-detail', args=[question_id]) which will lead to loss of query params and the user will be able to tell that his filtering is gone. so it is indeed user-visible. And I'm sure this way isn't putting the user at a security risk,

home/templates/home/question_detail.html
```diff
@@ -1,6 +1,7 @@
 {% extends "home/base.html" %}
```
rebeccaTubman, line 1: I tried to count that each div has a closing div and something doesn't add up correctly.

home/templates/home/question_detail.html
```diff
@@ -71,35 +72,35 @@ <h3> {{ question.title }} </h3>
...
+				<i class="bi bi-hand-thumbs-down-fill"></i>
+				{% else %}
+				<i class="bi bi-hand-thumbs-down"></i>
+			{% endif %}
+			  </a>
+			  - {{ answer.dislikes.count }}
+			  {% if not is_user_logged_in %}
+				<small><a href="{% url 'login' %}">Login</a> to like</small>
+			 {% endif %}
+			 </form>
```
rebeccaTubman, line 98: Why </form>? I can't find an opening form

home/tests/test_display_question_feature.py
```diff
@@ -127,3 +101,49 @@ def test_invalid_answersort_url(self, client):
             url = '/explore/question_1/?sortanswersby=BAD'
             response = client.get(url)
             assert response.status_code == 404
+
+    class TestThumbsRouting:
```
rebeccaTubman, line 105: LGTM
I see you tested all the status codes, view and updating the fields.

home/templates/home/question_detail.html
liornoy: You right, I should have deleted it. thank you

home/templates/home/question_detail.html
liornoy: I double-checked. it's probably the closing </form> tag

rebeccaTubman (approved):

idokk1 (approved):

liornoy:
The button appearance has changed, you can see the images in the linked issue.

danitLevi (approved):

ifireball (approved):

PR #116, opened 2021-05-13
Login redirect and display

### LINKED ISSUE #96: redirect login page
redirect login to previous page

idokk1:
- Add background to login page.
- redirect login page to last page.
- test for redirection.
fix #96

### DIFF
--- a/home/templates/home/login.html
+++ b/home/templates/home/login.html
@@ -3,10 +3,14 @@
 {% block title %}Login{% endblock %}
 
 {% block content %}
-  <h2>Login</h2>
-  <form method="post">
-    {% csrf_token %}
-    {{ form.as_p }}
-    <button type="submit">Login</button>
-  </form>
+<body class="main-bg">
+  <div class="mx-auto" style="width: 200px;">
+    <h2>Login</h2>
+    <form method="post">
+      {% csrf_token %}
+      {{ form.as_p }}
+      <button type="submit">Login</button>
+    </form>
+  <div>
+</body>
 {% endblock %}
--- a/home/tests/test_login.py
+++ b/home/tests/test_login.py
@@ -41,3 +41,8 @@ class TestLogin:
         response = client.get('/')
         assert "Login" not in str(response.content)
         assert "Logout" in str(response.content)
+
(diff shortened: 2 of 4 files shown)


home/tests/test_login.py
```diff
@@ -41,3 +41,8 @@ def test_authenticated_user_view(self, client, valid_user_details):
         response = client.get('/')
         assert "Login" not in str(response.content)
         assert "Logout" in str(response.content)
+
+    @pytest.mark.parametrize(('next'), [('about'), ('explore'), ('tags')])
+    def test_redirection_to_last_page_url(self, client, next, valid_user_details):
+        response = client.post('/login/?next=/{0}/'.format(next), valid_user_details)
+        assert response.url == '/{0}/'.format(next)
```
avivz450, line 48: I think you should add `assert response.status_code == 302`

idokk1: I have this assertion in `test_success_login_post_redirect`

avivz450 (approved):

shlifim/settings.py
```diff
@@ -70,6 +70,14 @@
     },
 ]
 
+TEMPLATE_CONTEXT_PROCESSORS = (
+    "django.core.context_processors.auth",
+    "django.core.context_processors.debug",
+    "django.core.context_processors.i18n",
+    "django.core.context_processors.media",
```
liornoy, line 77: what is "TEMPLATE_CONTEXT_PROCESSORS " used for? I can't find an explanation online.

idokk1: the first 4 lines are Django default. 
the 5th `django.core.context_processors.request`
create the ability to every request to know the previous path, to be able to return to the previous page after login.

rebeccaTubman (approved):

home/templates/home/login.html
```diff
@@ -3,10 +3,12 @@
 {% block title %}Login{% endblock %}
 
 {% block content %}
+<body class="main-bg">
```
danitLevi, line 6: @idokk1 
I recommend that you also add margins so that this screen will look better . 
You can do this by adding for example:
`class ="m-4" `

idokk1: added

ifireball (approved):

PR #118, opened 2021-05-13
Add sign up page 

### LINKED ISSUE #117: Create sign up page
need to create sign up page

idokk1:
Add sign up page, added tests for that.

fix #117

### DIFF
--- a/home/templates/home/signup.html
+++ b/home/templates/home/signup.html
@@ -1,0 +1,22 @@
+{% extends "home/base.html" %}
+{% load static %}
+{% block content %}
+<link rel="stylesheet" type="text/css" href="{% static 'home/CSS/new_question_style.css' %}">
+<body class="main-bg">
+  <div class="row pb-3">
+     <div class="column" style="width:25%"></div>
+        <div class="column" style = "width:59%" >
+          <div class="msg">
+            <center><h2>Sign up</h2></center>
+            <br><br>
+            <form method="post">
+              {% csrf_token %}
+              {{ form.as_p }}
+              <center><button class="submit-btn" type="submit">Sign up</button></center>
+            </form>
+        </div>
+      </div>
+    <div class="column"></div>
+  </div>
+</body>
+{% endblock %}
--- a/home/tests/test_login.py
+++ b/home/tests/test_login.py
@@ -4,10 +4,6 @@ import pytest
 @pytest.mark.django_db
 class TestLogin:
     @pytest.fixture
-    def valid_user_details(self):
-        return {"username": "Lior", "password": "LiorLior"}
-
-    @pytest.fixture
     def invalid_user_details(self):
         return {"username": "Lior12", "password": "LiorLior"}
 
--- a/home/tests/test_signup.py
+++ b/home/tests/test_signup.py
@@ -1,0 +1,64 @@
+import pytest
+from django.contrib.auth.models import User
+from home.models import Profile
+from pytest_django.asserts import assertTemplateUsed
+from home.forms import SignUpForm
+
+
+@pytest.mark.django_db
+class TestSignUp:
+
+    @pytest.fixture
+    def valid_signup_details(self):
+        return {'username': 'username', 'email': 'user@e.com', 'gender': 'U',
+                'password1': 'ido123123', 'password2': 'ido123123'}
+
+    def get_signup_page_status_code(self, client):
+        response = client.get('/signup')
+        assert response.status_code == 200
+
+    def test_valid_signup(self, valid_signup_details, client):
+        client.post('/signup/', data=valid_signup_details)
+        user = User.objects.filter(username=valid_signup_details['username'])
+        assert user is not None
+        profile = Profile.objects.filter(user=user)
+        assert profile is not None
+
+    @pytest.mark.parametrize("invalid_signup_details", [
+        # username cannot be empty
+        [{'username': '', 'email': 'valid@email.com', 'gender': 'U',
+         'password1': 'pw123123', 'password2': 'pw123123'}, 'username'],
+        # email invalid
+        [{'username': 'valid_username', 'email': 'a', 'gender': 'U',
+         'password1': 'pw123123', 'password2': 'pw123123'}, 'email'],
+        # password to short
+        [{'username': 'valid_username', 'email': 'valid@email.com', 'gender': 'U',
+         'password1': '123', 'password2': '123'}, 'password2'],
+        # passwords do not match
+        [{'username': 'valid_username2', 'email': 'valid@email.com', 'gender': 'U',
+         'password1': '000000000', 'password2': '123456789'}, 'password2']
+    ])
+    def test_invalid_signup(self, invalid_signup_details, client):
+        response = client.post('/signup/', data=invalid_signup_details[0])
+        user = User.objects.filter(username=invalid_signup_details[0]['username']).first()
+        assert invalid_signup_details[1] in response.context['form'].errors
+        assert user is None
+        assert response.status_code == 200
+
+    def test_valid_signup_redirection(self, valid_signup_details, client):
+        response = client.post('/signup/', data=valid_signup_details)
+        assert response.status_code == 302
+        assert response.url == '/'
+
+    def test_authenticated_user_view_signup(self, client, authenticated_user):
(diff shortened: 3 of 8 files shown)


home/tests/test_signup.py
```diff
@@ -0,0 +1,64 @@
+import pytest
+from django.contrib.auth.models import User
+from home.models import Profile
+
+
+@pytest.mark.django_db
+class TestSignUp:
+
```
avivz450, line 8: I think you should also add a test for the form and the template that are displayed.
You can see an example for that test in `test_question_form_and_template_displayed` method in test_insert_question_feature.py file

idokk1: good idea, added.

avivz450 (approved):

liornoy (approved):
Looks great, good tests coverage, and everything

rebeccaTubman (approved):
This PR looks great! Good job :)

home/templates/home/signup.html
```diff
@@ -0,0 +1,12 @@
+{% extends "home/base.html" %}
+
+{% block content %}
+<body class="main-bg">
```
danitLevi, line 4: @idokk1
I recommend that you also add margins so that this screen will look better and the form won't be too close to the sides.

idokk1: fixed at login page, here it's on purpose. there is a lot of text (like the minimum length of password ) and when it's on the center it's look worst

home/tests/test_signup.py
```diff
@@ -0,0 +1,71 @@
...
+        assert response.status_code == 200
+
+    def test_valid_signup(self, valid_signup_details, client):
+        client.post('/signup/', data=valid_signup_details)
+        try:
+            user = User.objects.filter(username=valid_signup_details['username'])
+            assert user is not None
+            profile = Profile.objects.filter(user=user)
+            assert profile is not None
+        except ValueError:
```
ifireball, line 27: You should avoid handling the exception like this, you are effectively hiding useful details from the test results report.

home/tests/test_signup.py
```diff
@@ -0,0 +1,71 @@
...
+         'password1': 'pw123123', 'password2': 'pw123123'},
+        # password to short
+        {'username': 'valid_username', 'email': 'valid@email.com', 'gender': 'U',
+         'password1': '123', 'password2': '123'},
+        # passwords do not match
+        {'username': 'valid_username2', 'email': 'valid@email.com', 'gender': 'U',
+         'password1': '000000000', 'password2': '123456789'}
+    ])
+    def test_invalid_signup(self, invalid_signup_details, client):
+        client.post('/signup/', data=invalid_signup_details)
```
ifireball, line 45: don't we expect to get some kind of an error/retry page?

How is the error presented to the user?

ifireball (changes requested):

home/tests/test_signup.py
idokk1: image

idokk1: don't really know how check it, this is built in functionality of Django.
tried to check if there is some response when trying to press signup with invalid data, but there isn't.

home/tests/test_signup.py
idokk1: fixed

danitLevi (approved):

home/tests/test_signup.py
idokk1: meanwhile checking that the page is not redirected (stay on signup page), and that the user is not created.

ifireball: The erros are included in the form object, see this example about how to render them manually to get an idea abut where they are:
https://docs.djangoproject.com/en/3.2/topics/forms/#rendering-form-error-messages

idokk1: Thanks, added tests.

rebeccaTubman (approved):
I see tests, no conflicts and a feature to help with the completeness.

home/tests/test_signup.py
```diff
@@ -0,0 +1,69 @@
...
+        assert response.status_code == 200
+
+    def test_authenticated_user_view_signup(self, client, authenticated_user):
+        response = client.get('/')
+        assert "Sign up" not in str(response.content)
+
+    def test_unauthenticated_user_view_signup(self, client, valid_user_details):
+        response = client.get('/')
+        assert "Sign up" in str(response.content)
+
```
liornoy, line 65: It looks like the valid_user_details argument is redundant here, doesn't it?

home/tests/test_signup.py
```diff
@@ -0,0 +1,69 @@
...
+    def test_invalid_signup_not_redirect(self, valid_signup_details, client):
+        invalid_user_details = valid_signup_details
+        invalid_user_details['username'] = ''
+        response = client.post('/signup/', data=invalid_user_details)
+        assert response.status_code == 200
+
+    def test_authenticated_user_view_signup(self, client, authenticated_user):
+        response = client.get('/')
+        assert "Sign up" not in str(response.content)
```
liornoy, line 60: It looks like the authenticated_userargument is redundant here, doesn't it?

home/tests/test_signup.py
```diff
@@ -0,0 +1,69 @@
...
+    def test_valid_signup_redirection(self, valid_signup_details, client):
+        response = client.post('/signup/', data=valid_signup_details)
+        assert response.status_code == 302
+        assert response.url == '/'
+
+    def test_invalid_signup_not_redirect(self, valid_signup_details, client):
+        invalid_user_details = valid_signup_details
+        invalid_user_details['username'] = ''
+        response = client.post('/signup/', data=invalid_user_details)
+        assert response.status_code == 200
```
liornoy, line 56: I think it's better to make the _invalid_user_details_ a fixture rather than what you did here, but anyways why not moving the response status code assert to the '_test_invalid_signup_' function where you check all sorts of invalid cases?

home/tests/test_login.py
```diff
@@ -3,10 +3,6 @@
 
 @pytest.mark.django_db
 class TestLogin:
-    @pytest.fixture
-    def valid_user_details(self):
-        return {"username": "Lior", "password": "LiorLior"}
-
     @pytest.fixture
     def invalid_user_details(self):
         return {"username": "Lior12", "password": "LiorLior"}
```
liornoy, line 8: Maybe move this also to conftest.py so you could use it

home/tests/test_signup.py
idokk1: yes it is

home/tests/test_signup.py
idokk1: no, the fixture authenticated_user login to an authenticated user . and then when I do client.get('/') i have a view of logged user. (thats why "sign up" is not in the content of the page)

home/tests/test_signup.py
idokk1: moved into test_invalid_signup , thanks

home/tests/test_login.py
idokk1: I have no need for this fixture in my PR (my invalid data includes more data, I did a parametrize instead)

home/templates/home/signup.html
```diff
@@ -0,0 +1,12 @@
...
+  <h2>Sign up</h2>
+  <form method="post">
+    {% csrf_token %}
+    {{ form.as_p }}
+    <button type="submit">Sign up</button>
```
danitLevi, line 9: @idokk1 
I think you can link the "login_style.css"  file to the "signup.html" file.
Then:
1. wrap the sign-in form with div with `class= "content"`
2. add in the button tag `class="login-btn"`
so this page will be designed same as the login page.

danitLevi (commented):
I think you can link here the login_style.css  and wrap the sign in form with div 
with `class= "content"`
and add the

danitLevi (approved):
LGTM

home/templates/home/signup.html
idokk1: added css like the add new question page.

idokk1:
hi @ifireball , this PR and #87 PR are without any conflicts, and I resolved all the issues , they are very important for the completeness of our app .

ifireball (approved):

PR #119, opened 2021-05-13
Added tags option to the 'insert a question' page + tests

### LINKED ISSUE #61: Add the tag feature when adding a question
Add the option to add up to 5 tags to a particular question

### LINKED ISSUE #109: Add failure messages once the question data is incorrect

avivz450:
- Added tags option to the 'insert a question' page + tests
- Added failure massages for invalid question + tests

fix #61 
fix #109

### DIFF
--- a/home/forms.py
+++ b/home/forms.py
@@ -5,9 +5,11 @@ from django.contrib.auth.models import User
 
 
 class QuestionForm(forms.ModelForm):
+    tags_ = forms.CharField(max_length=110, min_length=0, required=False)
+
     class Meta:
         model = Question
-        fields = ('title', 'subject', 'sub_subject', 'grade', 'book', 'book_page', 'content')
+        fields = ('title', 'subject', 'sub_subject', 'grade', 'book', 'book_page', 'tags_', 'content')
 
 
 class SignUpForm(UserCreationForm):
--- a/home/models.py
+++ b/home/models.py
@@ -202,6 +202,15 @@ class Tag(models.Model):
     def tags_feed(cls, search=''):
         return cls.objects.filter(tag_name__contains=search).order_by('tag_name')
 
+    @classmethod
+    def check_tag_array(cls, tags_array):
+        if(len(tags_array) > 5):
+            return False
+        for tag in tags_array:
+            if(len(tag) < 2 or len(tag) > 20):
+                return False
+        return True
+
 
 class Question_Tag(models.Model):
     question = models.ForeignKey(Question, on_delete=models.CASCADE, related_name='question_ID')
--- a/home/tests/test_insert_question_feature.py
+++ b/home/tests/test_insert_question_feature.py
@@ -140,24 +140,9 @@ class TestInsertQuestionFeature:
             ({'title': "Question in Math",
              'content': 'How much is it 1+1?', 'tags_': "", 'subject': 1}),
             # User entered all the required fields except grade
-            ({'title': "Question in Math", 'content': 'How much is it 1+1?',
-             'tags_': "", 'subject': 1, 'book_page': -23}),
-            # User entered all the required fields and a negative book page
-            ({'title': "Question in Math", 'content': 'How much is it 1+1?',
-             'tags_': "", 'subject': 1, 'book_page': 43343}),
-            # User entered all the required fields and to high book page
-            ({'title': "Question in Math", 'subject': 1, 'content': 'How much is it 1+1?',
-             'tags_': "aaaaaaaaaaaaaaaaaaaaaaaa", 'grade': Grade.GRADE7}),
-            # User entered all the required fields and invalid tags (more than 20 chars tag)
-            ({'title': "Question in Math", 'subject': 1, 'content': 'How much is it 1+1?',
-             'tags_': "a", 'grade': Grade.GRADE7}),
-            # User entered all the required fields and invalid tags (1 char tag)
-            ({'title': "Question in Math", 'subject': 1, 'content': 'How much is it 1+1?',
-             'tags_': "beyond_01, beyond_02, beyond_03, beyond_04, beyond_05, beoynd_06", 'grade': Grade.GRADE7}),
-            # User entered all the required fields and invalid tags (more than 5 tags)
         ])
         @pytest.mark.django_db
-        def test_post_invalid_question_with_client(self, client, invalid_data, authenticated_user):
+        def test_post_invalid_question_with_default_error_message(self, client, invalid_data, authenticated_user):
             response = client.post('/explore/new_question', data=invalid_data)
             assert response.status_code == 200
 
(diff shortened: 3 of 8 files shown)


idokk1:
I didn't found it in the code, 
if the user provide unvalid tags? there is a error message or you just not sending the form?

avivz450:
> I didn't found it in the code, 
> if the user provide unvalid tags? there is a error message or you just not sending the form?

You're right, I did not mention this in the PR description.  The plan is to add ab 'Alert' style messages in additional PR after this PR will be approved.  Right now, There are no messages in these cases

liornoy:
> I didn't found it in the code,
> if the user provides invalid tags? there is an error message or you just not sending the form?

What counts as an invalid tag?

idokk1:
> > I didn't found it in the code,
> > if the user provides invalid tags? there is an error message or you just not sending the form?
> 
> What counts as an invalid tag?

you can see the conditions in this PR in check_tag_array class method

home/views.py
```diff
@@ -39,9 +39,17 @@ def new_question(request):
...
-            questForm.save()
+            tags_array = [t.strip().lower() for t in questForm.cleaned_data['tags_'].split(',')]
+            if(not questForm.data['tags_']):
+                questForm = questForm.save(commit=False)
+                questForm.profile = request.profile
+                questForm.save()
+            elif(Tag.check_tag_array(tags_array)):
+                questForm = questForm.save(commit=False)
+                questForm.profile = request.profile
+                questForm.save()
```
danitLevi, line 50: @avivz450 
You have duplicate code here . 
the following code :
```
questForm = questForm.save(commit=False)   
questForm.profile = request.profile
questForm.save()
```
appears in both if and elif sections. 
I think you can execute this code and only after it write the if section

rebeccaTubman: Not exactly... we need to save the question only if:

- the tags field is empty - then we save only the question.
- the tags field isn't empty and valid - then we save the question and then connect the tags to the question object.

When we wrote it, we noticed the duplication but at the time couldn't exactly think of a better way. 
Do you see a way to remove the duplication and keep the logic?

idokk1: do this 3 lines 
`questForm = questForm.save(commit=False)   
questForm.profile = request.profile
questForm.save()`
without condition (or with condition of both) and then do the 2 other lines only if there is tags.

avivz450: We noticed the duplication of this code, but I have to keep it like that because I am going to add to the PR of the "Alert messages" an "else" option and it will display an error message.
Therefore, even if I will do what you said, in the next PR I will have to return the code to that way.

rebeccaTubman: I think we have to have a condition, otherwise we might save a question with unvalid tags.
`
tags_status = Tag.check_tag_array(tags_array)
if(not questForm.data['tags_'] or tags_status):
    questForm = questForm.save(commit=False)
    questForm.profile = request.profile
    questForm.save()            
    if(tags_status):
        questionInstane = Question.objects.get(id=questForm.id)
        questionInstane.add_tags_to_question(tags_array)
`
This is a way to avoid the duplication of the 2 lines but then there is a duplication in the if.
What do you think? @danitLevi @idokk1

danitLevi: @rebeccaTubman
looks good

avivz450: made that change

home/models.py
```diff
@@ -172,6 +172,15 @@ def __str__(self):
     def tags_feed(cls, search=''):
         return cls.objects.filter(tag_name__contains=search).order_by('tag_name')
 
+    @classmethod
```
rebeccaTubman, line 175: I think the test function for this method is missing.

home/views.py
```diff
@@ -39,9 +39,15 @@ def new_question(request):
...
-            questForm.save()
+            tags_array = [t.strip().lower() for t in questForm.cleaned_data['tags_'].split(',')]
+            tags_status = Tag.check_tag_array(tags_array)
+            if(not questForm.data['tags_'] or tags_status):
+                questForm = questForm.save(commit=False)
+                questForm.profile = request.profile
+                questForm.save()
+                if(tags_status):
+                    questionInstane = Question.objects.get(id=questForm.id)
+                    questionInstane.add_tags_to_question(tags_array)
```
rebeccaTubman, line 50: I'm glad it worked:)

home/models.py
avivz450: You right, I made a new file - `test_tag_model_functions.py` that will contain all of the tests that are related to his class models

rebeccaTubman: Thank you:)

home/tests/test_tag_model_functions.py
```diff
@@ -0,0 +1,47 @@
+import pytest
+from home.models import Tag
+
+
+class TestTagModelFunctions:
+    @pytest.mark.parametrize("valid_tags",
```
rebeccaTubman, line 6: Great! 
And thank you for re-organizing the tags tests

rebeccaTubman (approved):

home/forms.py
```diff
@@ -3,6 +3,8 @@
 
 
 class QuestionForm(forms.ModelForm):
+    tags_ = forms.CharField(max_length=110, min_length=0, required=False)
+
     class Meta:
         model = Question
-        fields = ('title', 'subject', 'sub_subject', 'grade', 'book', 'book_page', 'content')
+        fields = ('title', 'subject', 'sub_subject', 'grade', 'book', 'book_page', 'tags_', 'content')
```
liornoy, line 10: is it a typo? "tags_" ?

idokk1 (approved):

danitLevi (approved):
LGTM

home/forms.py
rebeccaTubman: No. Question model has a field tags, so the '_' was added to separate the model field and the form field.

home/views.py
```diff
@@ -42,10 +42,16 @@ def new_question(request):
     if request.method == 'POST':
         questForm = QuestionForm(request.POST)
         if questForm.is_valid():
-            questForm = questForm.save(commit=False)
-            questForm.profile = request.profile
-            questForm.save()
-            return redirect('question-detail', questForm.id)
+            tags_array = [t.strip().lower() for t in questForm.cleaned_data['tags_'].split(',')]
+            tags_status = Tag.check_tag_array(tags_array)
+            if(not questForm.data['tags_'] or tags_status):
```
ifireball, line 47: Not seeing updates to the tests of this view that reflect these code changes, also, assuming `check_tag_array` returns false, how do reflect that fat to the end user?

avivz450: You can see the updates to the tests that reflect these code changes in :
1. `test_tag_model_functions.py` - There is a test there that checks the new model function of Tag, `check_tag_array`.
2. `test_insert_question_feature.py` - I changed the parameterize of `valid_data` and `invalid_data` with some data of tags, as you can see in that file.
As a result, the tests in methods `test_add_valid_question` and `test_add_invalid_question` also check the insertion of the tags. 

Regarding `check_tag_array`, I thought about putting a message display mechanism in the case that `check_tag_array` is `False` in a separate PR.
Do you think it should be in this PR? (Because of time constraints)

@ifireball

home/tests/test_insert_question_feature.py
```diff
@@ -139,6 +157,42 @@ def test_post_question_without_title_with_client(self, client, authenticated_use
...
+            # User entered all the required fields and invalid tags (1 char tag)
+            (({'title': "Question in Math", 'subject': 1, 'content': 'How much is it 1+1?',
+             'tags_': "beyond_01, beyond_02, beyond_03, beyond_04, beyond_05, beoynd_06", 'grade': Grade.GRADE7}),
+             '''Invalid data.
+             You may enter up to 5 tags which are separated by ",".
+             Each tag must have between 2-20 characters.'''),
+            # User entered all the required fields and invalid tags (more than 5 tags)
+        ])
+        @pytest.mark.django_db
+        def test_post_invalid_question_with_custom_err_msg(self, client, invalid_data, error_msg, authenticated_user):
```
rebeccaTubman, line 187: Great job!

rebeccaTubman (approved):

ifireball (approved):

PR #121, opened 2021-05-15
Adding tags in explore page view

danitLevi:
**Adding tags in the explore page view** 
- Showing the tags of each question on the explore page 
- Testing questions tags appear on the explore page

### DIFF
--- a/home/tests/test_explore_page.py
+++ b/home/tests/test_explore_page.py
@@ -84,3 +84,11 @@ class TestExplorePage:
             url = f'/explore/?tag={invalid_tag_name}'
             response = client.get(url)
             assert response.status_code == 404
+
+        @pytest.mark.parametrize("tag_name", [
+            '#Pitagoras',
+            '#Bagrut_Exam'
+        ])
+        @pytest.mark.django_db
+        def test_explore_page_tags_view(self, explore_page_response, tag_name):
+            assert tag_name in str(explore_page_response.content)
--- a/home/templates/home/explore.html
+++ b/home/templates/home/explore.html
@@ -29,8 +29,17 @@
             <div class="col-3">
                 <a style="color:#6FB5EE" href="{% url 'question-detail' question.id %}"><b>{{ question.title }}</b></a>
             </div>
-            <div class="col">
-                <span>{{ question.subject }} ,{{ question.sub_subject }}</span>
+            <div class="col-3">
+                <div class="row">
+                    <span>{{ question.subject }} ,{{ question.sub_subject }}</span>
+                </div>
+                <div class="row mt-2">
+                        <div >
+                            {% for tag in question.tag_set.all %}
+                                <span class="badge badge-pill badge-secondary">#{{tag.tag_name}}</span>
+                            {% endfor %}
+                        </div>
+                </div>
             </div>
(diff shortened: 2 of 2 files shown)


idokk1:
LGTM

idokk1 (approved):

rebeccaTubman (approved):
Danit showed it to me and it's pretty!:)
I think it's a great addition to view of the explore page.

home/tests/test_explore_page.py
```diff
@@ -39,3 +39,11 @@ def test_explore_page_response(self, explore_page_response):
...
             char_content = explore_page_response.content.decode(explore_page_response.charset)
             assert '<a href="%s"' % reverse("new_question") in char_content
+
+        @pytest.mark.parametrize("tag_name", [
+            '#Pitagoras',
+            '#Bagrut_Exam'
+        ])
+        @pytest.mark.django_db
+        def test_explore_page_tags_view(self, explore_page_response, tag_name):
+            assert tag_name in str(explore_page_response.content)
```
avivz450, line 49: It looks like that you assume that you have at least one question with the tags "Pitagoras" and "Bagrut_Exam".
If it will be changed, for example by the "edit" option of the user or other option, that test will fail. 
I think these test should be like that :
1. add a question with tags that don't exist (check it before).
2. make this test with the new tags that you entered.

What do you think?

rebeccaTubman: I don't think we have an "edit" option, so I think it ok to leave this 2 tags.
I agree with aviv on adding maybe 2 new tags.

avivz450: I do not think it matters that we didn't implement that feature, equally the admin can delete the question for some reason, and then these tags will not exist.
The point is that I do not think we should rely on data that we actually have in the app when performing the tests, and therefore she should add a question with tags that do not exist in the app.

rebeccaTubman: Most of our test rely on data we have in the app.

danitLevi: @avivz450 
I test it this way because we don't allow the user to delete / edit questions (and not plan to do so ) therefore those tags should appear In the explore page. 
 Moreover, I agree with Rebecca that most of our tests rely on data we have in the app and therefore it seems that we can rely on them on the tests.
 @ifireball   I would be glad to hear your opinion regarding this issue.

ifireball: Not sure I understand @avivz450 point - the test data is there in order to be used in tests.

liornoy (approved):
Excellent addition

home/tests/test_explore_page.py
avivz450: This is irrelevant, I thought you were relying on the test data that was deleted after deleting db.sqlite3 and rebooting the virtual machine (using `vagrant halt`, `vagrant destroy`, and `vagrant up`).
Anyway, I realized this is not what I thought. I think that this test is great :)

avivz450 (approved):

danitLevi:
@ifireball
I resolved the conflicts in this PR and it is now ready to re-review

ifireball (approved):

PR #131, opened 2021-05-19
adjust question display to ckeditor

### LINKED ISSUE #132: remove tags from question display
before:
image

after:
image

idokk1:
strip html tags from question after replace the questions content to Rich Text Field (ckeditor).
you can see an example on the related issue.

- install bleach - html-sanitizer
- create new editor combined with bleach and ckedtior , to remove unwanted tags from question content

fix #132


danitLevi (approved):
Great fix

rebeccaTubman (approved):
Great job:)

home/templates/home/question_detail.html
```diff
@@ -39,7 +39,7 @@ <h3> {{ question.title }} </h3>
 		<b>by {{ question.profile }} </b>
 	  </div>
 		<div class="question-body">
-		  <p class="question-content">{{ question.content }}</p>
+		  <p class="question-content">{{ question.content | safe }}</p>
```
ifireball, line 42: This is scary shit... you've effectively helped your users to perform script injection on your pages... if you want to treat the question content as safe you must have some validation **on the server side** that makes sure it only contains a small subset of HTML that you can actually trust.

ifireball (changes requested):

home/templates/home/question_detail.html
idokk1: @ifireball added validation and tests for that

idokk1: after it will approve here I will do the same for #87

home/migrations/0019_alter_question_content.py
```diff
@@ -0,0 +1,19 @@
...
+
+    dependencies = [
+        ('home', '0018_add_tags_test_data'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='question',
+            name='content',
+            field=home.RichTextBleachField.RichTextBleachField(blank=True, null=True),
```
rebeccaTubman, line 17: Is it double on purpose?

idokk1: I dont see whats double?

rebeccaTubman: The `RichTextBleachField`

idokk1: yep the file name is same as the class

rebeccaTubman: Oh got it. Thanks:)

home/tests/test_insert_question_feature.py
```diff
@@ -154,3 +152,21 @@ def test_redirection_user_logged_out(self, client):
...
             assert response.url == '/login/?next=/explore/new_question'
+
+        @pytest.mark.django_db
+        def test_valid_tags_not_removed_from_question(self, client, authenticated_user):
+            data = {'title': "Question in Math",
+                    'subject': 1, 'content': '<p>a</p>',
+                    'grade': Grade.GRADE7}
+            client.post('/explore/new_question', data=data)
+            question = Question.objects.filter(title=data["title"]).first()
+            assert "<p>" in question.content
```
danitLevi, line 163: @idokk1 
I think we want to remove also the valid tags so the content will look clean to the user (who probably does not know how to program and don't know what those tags mean )

idokk1: the valid tags are remove in the HTML page (thats the |safe purpose) ,
what i prevent here is that unwanted tags will execute on |safe , like @ifireball example

ifireball (approved):

PR #143, opened 2021-05-27
Changed urls names

### LINKED ISSUE #142: Change URLS names, not paths
According to a StackOverflow post (below), URL names that are longer than one word should be separated with a dash.
in our case we had 'landingpage', 'explore-page' and 'new_question'. 
So I took it upon myself to even out all the names to be similar to 'explore-page'.
This change doesn't affect the user. the path address stays the same.

the post: https://stackoverflow.com/questions/31816624/naming-convention-for-django-url-templates-models-and-views

liornoy:
Changed the following:
'landingpage' -> 'landing-page'
'new_question' -> 'new-question'

The change is in the urls.py file, but also in the HTML files that call the URLs,
and also in one test that used an old name.

This is so the URLs names will be consistent and in format
fix #142

### DIFF
--- a/home/templates/home/base.html
+++ b/home/templates/home/base.html
@@ -26,7 +26,7 @@
 	<body>
 	<nav class="navbar navbar-expand-lg navbar-light">
 	  <div class="container-fluid">
-		<a class="nav-link active" href="{% url 'landingpage' %}">
+		<a class="nav-link active" href="{% url 'landing-page' %}">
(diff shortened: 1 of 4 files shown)


avivz450 (approved):

danitLevi (approved):

danitLevi:
LGTM

idokk1 (approved):

PR #146, opened 2021-05-28
Add title and content fields to messages-system

### LINKED ISSUE #145: Upgrade messages system
Currently, if you want to use the messages system, the word "ERROR:" is being printed no matter what.
`<strong>ERROR: </strong> {{ message }}` - base.html
Need to add a way to allow different kinds of messages like alert, and success.
therefore, 'title' and 'message_content' fields were added to the message so it's more flexible.

* I need this fix for using it in the delete question option, where I want a confirmation success message to appear

liornoy:
This way we can show messages other than ERROR:
fix #145

### DIFF
--- a/home/views.py
+++ b/home/views.py
@@ -87,9 +87,10 @@ def new_question(request):
                 tags_error_msg = '''Invalid data.
              You may enter up to 5 tags which are separated by ",".
              Each tag must have between 2-20 characters.'''
-                messages.error(request, tags_error_msg)
+                messages.error(request, {'title': 'ERROR:', 'message_content': tags_error_msg})
         else:
-            messages.error(request, 'Invalid data. Book number must be between 1-999.')
+            messages.error(request, {'title': 'ERROR:',
+                           'message_content': 'Invalid data. Book number must be between 1-999.'})
     return render(request, 'home/questions/new_question.html', {'form': form, 'title': 'New Question'})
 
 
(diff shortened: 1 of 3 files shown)


avivz450:
Not sure that I understood the contribution in this PR

liornoy:
> Not sure that I understood the contribution in this PR @liornoy

The message system can be used in different situations different from Errors. and right now it shows the text "ERROR" no matter what. I am interested in using messages for giving confirmation to the user for deleting his question, and I don't want "ERROR" in my message. so I made it that you can edit the word in bold to fit other cases. in my case I will print "SUCCESS!" instead of "ERROR"

home/views.py
```diff
@@ -87,9 +87,10 @@ def new_question(request):
                 tags_error_msg = '''Invalid data.
              You may enter up to 5 tags which are separated by ",".
              Each tag must have between 2-20 characters.'''
-                messages.error(request, tags_error_msg)
+                messages.error(request, {'title': 'ERROR:', 'message_content': tags_error_msg})
```
avivz450, line 90: Ok, great addition!
I think you should add a test for the title as well.

home/views.py
```diff
@@ -87,9 +87,10 @@ def new_question(request):
...
         else:
-            messages.error(request, 'Invalid data. Book number must be between 1-999.')
+            messages.error(request, {'title': 'ERROR:',
```
avivz450, line 92: Also here

avivz450 (changes requested):

home/views.py
liornoy: ok will do

home/views.py
liornoy: Added a check for the title to the Insert new question tests, and it covers both of the things you mentioned

avivz450 (approved):

PR #149, opened 2021-05-29
Added link from tags to explore page

### LINKED ISSUE #148: Add links to the tags badges
We display tags for each question in the explore page and in the display question page.
The tags are inside badges. I think we should add links so if a user clicks a badge of some tag, it will
send him to the explore page filtered by this tag.

In order to do so, I will add "a" tag inside the badge and remove the style of it except the mouse icon that will change as if it's a button.

liornoy:
In every place we display tags, (explore page and display question
page) I added an option to click on the tag's badge and it will redirect
to the explore page filtered to show all questions with this tag.

No tests were added because this link's URL has already been
tested during the tags page feature.
fix #148

### DIFF
--- a/home/static/home/CSS/base_style.css
+++ b/home/static/home/CSS/base_style.css
@@ -40,3 +40,14 @@ body {
 	padding-left:5px;
 	padding-right:5px;
 }
+
+
+a.nostyle:link {
+    text-decoration: inherit;
+    color: inherit;
+}
+
+a.nostyle:visited {
+    text-decoration: inherit;
+    color: inherit;
+}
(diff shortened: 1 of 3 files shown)


home/static/home/CSS/base_style.css
```diff
@@ -40,3 +40,14 @@ body {
...
+
+a.nostyle:link {
+    text-decoration: inherit;
+    color: inherit;
+}
+
+a.nostyle:visited {
+    text-decoration: inherit;
+    color: inherit;
+}
```
liornoy, line 53: The style was added to base_style.css because I want to use this style in different pages, so placing it here will allow me to use it from every where

avivz450 (approved):

PR #151, opened 2021-05-29
Change client.login to client.force_login

### LINKED ISSUE #150: Switch to force_login to reduce test duration
Currently, with 151 tests, it takes 44.58s to run the tests.
By switching to the `force_login ` method we can reduce the time significantly. 
image

liornoy:
Reduces Testing time from 44.58s to 17.50s.

force_login was recommended by Barak, and it does the same as a login but without the encryption/authentication logic behind the scenes so it faster.

fix #150


avivz450 (approved):
Great addition!

PR #152, opened 2021-05-29
User page

### LINKED ISSUE #147: User-page feature
User or Profile page will show all the questions and answers a user has made,
with links to the question and answers.

The URL for this page will be "users/<user_id>"
and you get to this page from links in explore page and display-question page
every place where it shows a profile that wrote a question or answer will be a link to this user's page.

liornoy:
A user page that shows all the users,
and an individual user page for each user, that shows:
1. username
2. join date
3. num of answers - and the answers
4. num of questions - and the questions
5. gender
added also links to everywhere a user name is shown
You can also get to your own profile page by clicking you username in the right side of the navbar

fix #147

### DIFF
--- a/home/models.py
+++ b/home/models.py
@@ -44,6 +44,10 @@ class Profile(models.Model):
         profile.save()
         return profile
 
+    @classmethod
+    def profiles_feed(cls):
+        return cls.objects.all()
+
 
 class Subject(models.Model):
     subject_name = models.CharField(max_length=100, unique=True)
@@ -137,6 +141,10 @@ class Question(models.Model):
     def get_answers_num(self):
         return self.answer_set.count()
 
+    @classmethod
+    def get_all_user_questions(cls, profile):
+        return cls.objects.all().filter(profile=profile)
+
 
 class Answer(models.Model):
     profile = models.ForeignKey(Profile, on_delete=models.CASCADE)
@@ -197,6 +205,10 @@ class Answer(models.Model):
     @classmethod
     def get_answers_by_date(cls):
         return cls.objects.order_by('-publish_date')
+
+    @classmethod
+    def get_all_user_answers(cls, profile):
+        return cls.objects.all().filter(profile=profile)
 
 
 class Tag(models.Model):
--- a/home/templates/home/user_page.html
+++ b/home/templates/home/user_page.html
@@ -1,0 +1,30 @@
+{% extends "home/base.html" %}
+{% load static %}
+{% block content %}
+<link rel="stylesheet" type="text/css" href="{% static 'home/CSS/user_page.css' %}">
+<body class="main-bg">
+<main role="main" class="container">
+ <h1 class="title">{{  profile.user.username }} Profile page</h1>
+ <p> Date joined: {{ profile.user.date_joined | date:"Y-F-d" }}</p>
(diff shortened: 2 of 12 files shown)


home/templates/home/user_page.html
```diff
@@ -0,0 +1,26 @@
+{% extends "home/base.html" %}
+{% load static %}
+{% block content %}
+<link rel="stylesheet" type="text/css" href="{% static 'home/CSS/user_page.css' %}">
+<body class="main-bg">
+<main role="main" class="container">
+ <h1 class="title">{{ profile }} Profile page</h1>
```
avivz450, line 7: If the profile name is "Lior", we get from this line: "Lior Profile page".
The reason of that is because __str__ is returning `'{self.user.username}'.format(self=self)`.
The problem is that if `__str__` will be changed in the future, we will get a different output here.
Therefore, I think it should be {{ profile.user.username }} instead of {{ profile }}.

home/templates/home/user_page.html
```diff
@@ -0,0 +1,26 @@
...
+		<a class="questions" href="{% url 'question-detail' question.pk  %}">{{question}}</a>
+		<br/>
+	{% endfor %}
+  </div>
+  <div class="column right-col border border-dark">
+  <h2>User Answers</h2>
+  {% for answer in user_answers %}
+		<a class="answers" href="{% url 'question-detail' answer.question.pk  %}">Answered {{answer.question}}</a>
+		<br/>
+	{% endfor %}
```
avivz450, line 21: Maybe we should put more data on the profile's page besides the user's answers and questions.
I thought about adding as well: num of answers, num of questions, registration date, and email.
What do you think?

home/models.py
```diff
@@ -137,6 +141,10 @@ def get_question_title(self):
     def get_answers_num(self):
         return self.answer_set.count()
 
+    @classmethod
+    def get_all_user_questions(cls, profile):
+        return cls.objects. all().filter(profile=profile)
```
avivz450, line 146: there is a space here

home/templates/home/user_page.html
liornoy: I'm down to add num of.. and registration date, email is something that should be kept privately IMO, and maybe I can add the gender with an adjacent icon if we already have this info.

home/templates/home/user_page.html
liornoy: great comment

home/templates/home/user_page.html
avivz450: Ok, it is possible at right now not to display this data and in the future maybe we will add an option to ask the user which data he prefers to be displayed and which not.

home/models.py
liornoy: thanks fixed

avivz450 (approved):
Please edit the description of the PR with the new data that is displayed on the user page before merging

liornoy:
@avivz450 I added a link from the username in the top right in the navbar to the profile page. and added reverse function to the tests. is it cool with you?

avivz450:
> @avivz450 I added a link from the username in the top right in the navbar to the profile page. and added reverse function to the tests. is it cool with you?

yep
Please try next time to make the PRs smaller so it will be easier to review (and with fewer commits).

PR #158, opened 2021-06-07
Fix no answer content - bug fix

### LINKED ISSUE #155: Post answer allow posting empty answer with no content

liornoy:
fix #155

### DIFF
--- a/home/views.py
+++ b/home/views.py
@@ -36,9 +36,10 @@ def display_question_page(request, pk):
             if (request.profile is None):
                 return HttpResponse('Unauthorized', status=401)
             else:
-                form.instance.profile = request.profile
-                form.instance.question = question
-                form.save()
+                if form.cleaned_data['content'] != "":
+                    form.instance.profile = request.profile
+                    form.instance.question = question
+                    form.save()
             return redirect(reverse("question-detail", kwargs={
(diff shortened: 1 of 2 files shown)


home/views.py
```diff
@@ -36,9 +36,10 @@ def display_question_page(request, pk):
             if (request.profile is None):
                 return HttpResponse('Unauthorized', status=401)
             else:
-                form.instance.profile = request.profile
-                form.instance.question = question
-                form.save()
+                if form.cleaned_data['content'] != "":
+                    form.instance.profile = request.profile
+                    form.instance.question = question
+                    form.save()
```
avivz450, line 42: I don't think it's a good way to solve that bug.
I think that we should alter the `content` field in the `Answer` model instead.
I mean, I think that instead of `content = RichTextBleachField(blank=True, null=True)` in the `Answer` model, it should be `content = RichTextBleachField(default = None)`.
After making a migration file for that + migrate, It should work the way we want.

avivz450 (changes requested):

home/views.py
avivz450: By the way, after doing that you can delete `if form.cleaned_data['content'] != "":`

PR #160, opened 2021-06-07
Add the use of reverse method in tests

### LINKED ISSUE #159: Use reverse in tests
In the tests, we use client.get(url) a lot, and not using the reverse method in the URL, which will be hard to maintain if you will change the URLs in the future.

liornoy:
Switched the code so that tests will use reverse instead of hard-wired URLs, so if we change the URLs in the future, it will automatically change in the tests.

edit: Also moved @pytest.mark.django_db tag above the class header, so it won't repeat every test header

fix #159

### DIFF
--- a/home/tests/test_about_page.py
+++ b/home/tests/test_about_page.py
@@ -2,8 +2,8 @@ import pytest
 from django.urls import reverse
 
 
+@pytest.mark.django_db
 class TestAboutPage:
-    @pytest.mark.django_db
     @pytest.fixture
     def about_page_response(self, client):
         url = reverse("about")
@@ -13,7 +13,6 @@ class TestAboutPage:
     def test_about_page_url(self, about_page_response):
         assert about_page_response.status_code == 200
 
-    @pytest.mark.django_db
     @pytest.mark.parametrize(
         ("link"),
         [
--- a/home/tests/test_add_answer.py
+++ b/home/tests/test_add_answer.py
@@ -1,27 +1,41 @@
 from home.models import Answer
+from django.urls import reverse
 import pytest
 
 
 @pytest.mark.django_db
 class TestAddAnswer:
     @pytest.fixture
+    def question_details_url(self, question_test_data):
+        return reverse("question-detail", args=[question_test_data.id])
+
+    @pytest.fixture
     def comment_content(self):
         return {"content": "test"}
 
     def test_post_redirect(
-        self, client, question_test_data, authenticated_user, comment_content
+        self, client, question_details_url, authenticated_user, comment_content
     ):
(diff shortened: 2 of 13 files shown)


avivz450 (approved):
Very nice!

liornoy:
@avivz450 I call reverse billion times in the tests and that's a lot of function calls.
How can I optimize it?

avivz450:
> @avivz450 I call reverse billion times in the tests and that's a lot of function calls.
> How can I optimize it?
It looks like you're calling reverse the same amount of times you called the URLs before this PR, why do you think this is a problem?

liornoy:
> > @avivz450 I call reverse billion times in the tests and that's a lot of function calls.
> > How can I optimize it?
> 
> It looks like you're calling reverse the same amount of times you called the URLs before this PR, why do you think this is a problem?

Because previously it was just client.get("some-url") and now it's client.get(reverse("url-name"))
I mean before it was a string and now it is a function call. and it's a problem because there are tests where I call 5+ times to the same function that returns the same result which I can save the return value and use instead.
What I think could be the best option is that when I run the Pytest then I would call once to all the reverse of each URL, save it in global variables like `landing_page_url.`, and use it in all the needed tests. And I tried to do it in the conftest.py file but it doesn't work... So if you have an idea to how to make it global to all tests, and if not maybe I should do this in each test, I mean in the beginning of the test to save all the URL paths to local vars by calling reverse and use the vars instead calling multiple time to the function.

avivz450:
> > > @avivz450 I call reverse billion times in the tests and that's a lot of function calls.
> > > How can I optimize it?
> > 
> > 
> > It looks like you're calling reverse the same amount of times you called the URLs before this PR, why do you think this is a problem?
> 
> Because previously it was just client.get("some-url") and now it's client.get(reverse("url-name"))
> I mean before it was a string and now it is a function call. and it's a problem because there are tests where I call 5+ times to the same function that returns the same result which I can save the return value and use instead.
> What I think could be the best option is that when I run the Pytest then I would call once to all the reverse of each URL, save it in global variables like `landing_page_url.`, and use it in all the needed tests. And I tried to do it in the conftest.py file but it doesn't work... So if you have an idea to how to make it global to all tests, and if not maybe I should do this in each test, I mean in the beginning of the test to save all the URL paths to local vars by calling reverse and use the vars instead calling multiple time to the function.

I don't know another way to make these variables global.
I thought maybe putting in the same class tests where the same reverse call is made, and save the output of the same reverse call in a variable ,but it's not sure there is a logical connection between them.
I think what you suggested is good enough.

avivz450 (approved):

PR #161, opened 2021-06-08
Added black settings and ci

avivz450:
Installed "black" and added workflow for CI

Motivation :
By using this tool, we will not have to manually edit the code to fit the flake8 format

For formatting files locally, run the command -
"pipenv run python -m black ."


liornoy (approved):
Good addition. I think we can merge.
shouldn't we combine all the 3 yml files in .github workflows into 1 file that will do everything?
because I think right now github run 3 times the fedora os and 3 times intalls pipenv and git and runs each yml code.

avivz450:
> Good addition. I think we can merge.
> shouldn't we combine all the 3 yml files in .github workflows into 1 file that will do everything?
> because I think right now github run 3 times the fedora os and 3 times intalls pipenv and git and runs each yml code.

No problem, I think it should work and be more efficient.
If you want, you can do it in another PR.

PR #163, opened 2021-06-18
Fix no-content in answer bug

liornoy:
Changed the attributes of RichTextBleachField, from blank=True, null=True
to - default = ''

fix #158


avivz450 (approved):

home/migrations/0020_alter_answer_content.py
```diff
@@ -14,6 +14,6 @@ class Migration(migrations.Migration):
         migrations.AlterField(
             model_name="answer",
             name="content",
-            field=home.RichTextBleachField.RichTextBleachField(blank=True, null=True),
+            field=home.RichTextBleachField.RichTextBleachField(default=""),
```
avivz450, line 17: I think you should also add a test of trying to post an empty answer

avivz450 (changes requested):

PR #175, opened 2021-06-22
Split shlifim into users app and home app

### LINKED ISSUE #173: Split the project to home and users apps
After creating the new app users, the following things should get done:
Move to users app the following: views, models, tests, forms, templates, URLs

liornoy:
This is a big PR, it can't be smaller because once you split the project into 2 apps, you can't
move them piece by piece, but all in once.
I made sure the PR is split into small commits, each one handles one thing.
From the user point of view, this PR only changes the URL written in the address bar
but other than that nothing changed, and this PR will help us the developers by making the code
more organized and easy to workaround

fix #173


avivz450 (commented):
What about the CSS files? Don't you need to move these static files to the users folder as well?

users/tests/test_signup.py
```diff
@@ -106,4 +106,4 @@ def test_unauthenticated_user_view_signup(self, client):
     def test_signup_form_and_template_displayed(self, client):
```
avivz450, line 106: Line 29 shouldn't be `client.post(self.signup_url, data=valid_signup_details)`?

avivz450 (commented):
Great documentation!

users/admin.py
```diff
@@ -1,3 +1,5 @@
-# from django.contrib import admin
+from django.contrib import admin
+from .models import Profile
 
-# Register your models here.
+
+admin.site.register(Profile)
```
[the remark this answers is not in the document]
liornoy: As you can see in the picture below, the Profile had moved to a different area 
image

users/tests/test_signup.py
liornoy: Thanks! didn't notice it

avivz450 (approved):

PR #184, opened 2021-07-03
Fix post question bug

### LINKED ISSUE #156: Post question allow sub-subject not matching subject

liornoy:
The bug is that you could upload a new question with a sub_subject not matching the subject.
The update is mainly in the NewQuestionForm and the new_question view.
Also added 2 new tests.
fix #156

### DIFF
--- a/home/forms.py
+++ b/home/forms.py
@@ -19,6 +19,14 @@ class QuestionForm(forms.ModelForm):
             "content",
         )
 
+    def clean_sub_subject(self):
+        subject = self.cleaned_data.get("subject")
+        sub_subject = self.cleaned_data.get("sub_subject")
+        if sub_subject and sub_subject not in subject.sub_subject_set.all():
+            raise forms.ValidationError("The sub-subject doesn't match the subject")
+        else:
+            return sub_subject
+
 
 class CommentForm(forms.ModelForm):
     content = RichTextBleachField(default="")
--- a/home/tests/test_insert_question_feature.py
+++ b/home/tests/test_insert_question_feature.py
@@ -330,6 +330,16 @@ class TestInsertQuestionFeature:
                     }
                 ),
                 # User entered all the required fields and a valid book page
+                (
+                    {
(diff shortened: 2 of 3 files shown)


home/forms.py
```diff
@@ -19,6 +19,14 @@ class Meta:
             "content",
         )
 
+    def clean_sub_subject(self):
+        subject = self.cleaned_data.get("subject")
+        sub_subject = self.cleaned_data.get("sub_subject")
+        if sub_subject and sub_subject not in subject.sub_subject_set.all():
```
avivz450, line 25: I'm not sure I understood the check here.
I don't see the `sub_subject_set` field in the `Subject` model, How do you check that a specific sub-subject is related to a certain subject?

home/tests/test_insert_question_feature.py
```diff
@@ -465,6 +475,19 @@ def test_post_question_without_title_with_client(
...
+                    (
+                        {
+                            "title": "Question in Math",
+                            "subject": 1,
+                            "sub_subject": 10,
+                            "content": "How much is it 1+1?",
+                            "grade": Grade.GRADE7,
+                        }
+                    ),
+                    """The sub-subject doesn't match the subject""",
```
avivz450, line 488: Is there a test of matching sub-subject to a certain subject?

liornoy: there isn't, I'll add

home/forms.py
liornoy: it's because sub_sucject has subject field as ForeignKey so `sub_subject_set ` is automatic

avivz450: Just to make sure I understand, `sbj.sub_subject_set.all()` returns a list of all of the `sub_subject`s that has `sbj` as foreign key, right?

avivz450 (approved):

home/forms.py
liornoy: yes

PR #188, opened 2021-07-11
Fix client.get calls, to use param dictionary

### LINKED ISSUE #187: Fix the client.get calls in the tests to use dictionary for query param
After reading the Doc in: https://docs.djangoproject.com/en/3.2/topics/testing/tools/#django.test.Client.get
I refactored the client.get calls where we have query params. previously we concatenated strings and now changed it to use a dictionary.

liornoy:
In the official Django doc for the client.get function, it mentioned
that if you want to make a request with query param you can
send key-value pairs (dictionary)

fix #187

### DIFF
--- a/home/tests/test_insert_question_feature.py
+++ b/home/tests/test_insert_question_feature.py
@@ -381,7 +381,7 @@ class TestInsertQuestionFeature:
                         "content": "How much is it 1+1?",
                         "tags_": "",
                         "subject": 1,
-                        "sub_subject": 10
+                        "sub_subject": 10,
                     }
                 )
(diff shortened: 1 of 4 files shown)


home/tests/test_insert_question_feature.py
```diff
@@ -381,7 +381,7 @@ def test_post_valid_question_with_client(
                         "content": "How much is it 1+1?",
                         "tags_": "",
                         "subject": 1,
-                        "sub_subject": 10
+                        "sub_subject": 10,
```
liornoy, line 384: this was added by black automatically

avivz450 (approved):

PR #189, opened 2021-07-11
Create generator function for display question

### LINKED ISSUE #179: Change answers_tuples data structure in display-question
Barak suggested: make it return an iterator or be a generator method instead of returning a full list in memory
Needs to research this topic

liornoy:
This is an optimization change recommended by Barak.
It lowers the memory usage by using a lazy-iterator.
fix #179 
The changes are in the models.py and display question test file

### DIFF
--- a/home/models.py
+++ b/home/models.py
@@ -162,16 +162,13 @@ class Answer(models.Model):
     @classmethod
     def get_answers_tuples(cls, question, sort_answer_by, profile):
         answers = question.get_answers_feed(sort_answer_by)
-        answers_tuples = []
         for answer in answers:
-            answers_tuples.append(
-                (
+            answers_tuples = (
                     answer,
                     answer.profile_liked(profile),
                     answer.profile_disliked(profile),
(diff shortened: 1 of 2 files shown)


avivz450:
I could not find information on how there is more reduced memory usage as you mentioned.
I will continue to search and check, I would love for you to update me on how the improvement here is being made once you will know.

liornoy:
> I could not find information on how there is less memory usage in these cases as you mentioned.
> I will continue to search and check, I would love for you to update me on how the improvement here is being made once you will know.

@avivz450  - Barak clarified this to me, the improvement of the memory utilization is noticeable mostly on a deployment env. where for example imagine the Django server is handling with thousands of users requests, and if at once many users will request this view, which will trigger the function and every function call will generate a list full of answers, then you might reach the max capacity of the memory available. When on the other end, the generator function will only return 1 answer at a time, so in this case the generator function is better.

avivz450 (approved):

